PR #51, opened 2024-01-16
Add first cookbook example for function calling

devashishtyagi:
Adding an example that is based off retrieval example from langchain.

Example run - https://colab.research.google.com/drive/1naculvcG0RrZhglsdRBUz_fXHyVDOtDN?usp=sharing


pgarbacki (approved):

recipes/examples/fireworks_functions_qa.ipynb
```diff
@@ -0,0 +1,457 @@
+{
```
pgarbacki, line 1: could you move it to `recipes/examples/function_calling/fireworks_functions_qa.ipynb`?

PR #52, opened 2024-01-16
reorg directory to make examples top-level

devashishtyagi:
Make example top level because currently the structure makes little sense. Recipes contains example workflows to interact with GenAI infrastructure. It doesn't contain full fledged applications. In order to create a separate space for contributing GenAI examples we make it a top level folder along with documentation.


pgarbacki (approved):

Hecate0821:

PR #76, opened 2024-02-29
Update MongoDB tutorials round 2

benjibc:


devashishtyagi:
> not batching the embedding lookups, and just have a for loop that goes through all the documents and embed them manually
and we are not batching... instead we just have

Additionally it's nice to include a open in colab link like this - https://github.com/fw-ai/cookbook/blob/main/examples/function_calling/fireworks_functions_information_extraction.ipynb

devashishtyagi (approved):

PR #85, opened 2024-08-14
Fireworks x LangGraph

shiv248:
Hello! 
I was at the Agents and Compound AI Hackathon this past weekend and saw there wasn't an up to date example of Fireworks x LangGraph tool calling. I went ahead and wrote one up to use.

implementation of LangGraph tool calling with `langgraph-fireworks` library and fireworks interface model using `FireFunction-v2`.


pgarbacki (approved):
Thanks for contributing!

PR #124, opened 2025-06-18
Feat: add function calling to voice agents

RobertoBarrosoLuque:
- Update voice agent example with function calling
- Update readme with description on how to use function calling

### DIFF
--- a/learn/voice-agent/main.py
+++ b/learn/voice-agent/main.py
@@ -222,7 +222,7 @@ class VoiceAgent:
                     "answer": {
                         "system_prompt": PROMPT.strip(),
                         "max_tokens": 150,  # Shorter responses for better flow
-                        "temperature": 0.7,  # More natural but consistent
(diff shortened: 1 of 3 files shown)


learn/voice-agent/main.py
```diff
@@ -15,61 +17,177 @@
 CHUNK_SIZE = 1024
 TTS_SAMPLE_RATE = 44100
 
-ENDPOINT = "wss://audio-agent.link.fireworks.ai/v1/audio/agent"
+TODAY = dt.date.today().strftime("%Y-%m-%d")
```
dmitry-mli, line 20: May need time also. Sample: "my appointment is3 hours from now, i want to cancel it"

dmitry-mli (approved):

learn/voice-agent/main.py
RobertoBarrosoLuque: Updated!

PR #173, opened 2026-03-05
multiple improvements

mayinghan:
1. add sft example with qwen3 8b default in the driver script
2. add rft trajectory dumping (jsonl)
3. add more proper cleanup (on all exit signals) to prevent resource leakage
4. remove hotload and deployment from sft loop

### DIFF
--- a/training/examples/deepmath_rl/train_deepmath.py
+++ b/training/examples/deepmath_rl/train_deepmath.py
@@ -15,6 +15,8 @@ import time
 from dataclasses import dataclass, field
 from typing import cast
 
+from dotenv import load_dotenv
+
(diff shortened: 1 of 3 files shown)


training/examples/sft/train_sft.py
```diff
@@ -0,0 +1,100 @@
...
+
+from __future__ import annotations
+
+import argparse
+import os
+import sys
+import logging
+import signal
+
+_SRC = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", "..", ".."))
```
RobertoBarrosoLuque: I dont think we need this. If clients properly use the setup instruction the pyproject.toml should address this

RobertoBarrosoLuque (approved):

PR #188, opened 2026-03-10
fix: eagerly load tokenizer once instead of per-rollout row

benjibc:
## Summary

- The tokenizer and completions clients were being created inside `process_row()`, meaning every rollout row triggered a fresh `AutoTokenizer.from_pretrained()` call to HuggingFace
- On CI runners (or any shared IP without `HF_TOKEN`), this causes 429 rate limiting after the first few calls, failing all subsequent rollouts
- Move tokenizer loading, `tool_call_parser`, and client creation outside `process_row()` so they are initialized once per `__call__` invocation and shared across all rows

### Changes in `frozen_lake_rollout.py`

- `_get_hf_tokenizer()` called once at the top of `FrozenLakeToolRolloutProcessor.__call__`
- `FireworksV1ImageCompletionsClient` / `FireworksV1CompletionsClient` created once and shared
- `build_frozen_lake_tool_call_parser` called once with the shared tokenizer
- `FireworksV1ImageCompletionsClient.__init__` eagerly loads tokenizer instead of lazy-loading per request

### Related

- Fireworks PR: https://github.com/fw-ai/fireworks/pull/19132 (RLOR smoke test references this branch)

## Test plan

- [x] RLOR smoke test no longer hits HuggingFace 429 rate limits during rollouts
- [ ] Verify green CI run end-to-end

### DIFF
--- a/training/examples/frozen_lake/frozen_lake_rollout.py
+++ b/training/examples/frozen_lake/frozen_lake_rollout.py
@@ -255,7 +255,7 @@ class FireworksV1ImageCompletionsClient:
         self.enable_thinking = enable_thinking
         self.tool_call_parser = tool_call_parser
         self.default_tools = default_tools or []
-        self._tokenizer = None
+        self._tokenizer = _get_hf_tokenizer(self.tokenizer_name_or_path)
         self._client = None
 
     async def close(self) -> None:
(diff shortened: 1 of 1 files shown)


Hecate0821:

PR #206, opened 2026-03-13
fix(frozen-lake): harden gateway-backed smoke path

benjibc:
## Summary

- Fall back to trainer-gateway health checks when pre-created FrozenLake trainer job lookups fail, so smoke runs can still attach to already-running jobs instead of dying on follow-up control-plane calls
- Use a single `inference_api_base` for both rollout traffic and deployment readiness probes, and run readiness through the same `AsyncFireworks` client path the recipe uses for inference
- **Add `lora_mode` to `WeightSyncer` instantiation** in both `rl_loop.py` and `train_frozen_lake.py` — LoRA adapter checkpoints are always complete (not diffs), so incremental/delta hotload metadata must be disabled to avoid crashing the serving container with a 503

## Test plan

- [x] End-to-end RLOR smoke test (LoRA + full-param + Kimi B300 80k) all 3 jobs passing in CI

## Related

- Pairs with [fw-ai/fireworks#19411](https://github.com/fw-ai/fireworks/pull/19411) (trainer-side LoRA DCP + CI smoke test)
- SDK `lora_mode` field added in `fireworks-ai-python` `weight_syncer.py`

### DIFF
--- a/training/examples/frozen_lake/train_frozen_lake.py
+++ b/training/examples/frozen_lake/train_frozen_lake.py
@@ -32,6 +32,7 @@ _SRC = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..",
 if _SRC not in sys.path:
     sys.path.insert(0, _SRC)
 
+from fireworks import AsyncFireworks
 from eval_protocol.models import EvaluationRow, InputMetadata, Message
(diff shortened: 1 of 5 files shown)


Hecate0821:

PR #207, opened 2026-03-14
add reinforce algo

websterbei:
Adding reinforce with optional KL penalty

### DIFF
--- a/training/recipes/rl_loop.py
+++ b/training/recipes/rl_loop.py
@@ -790,30 +790,5 @@ def main(
 
 
 if __name__ == "__main__":
-    from dotenv import load_dotenv
-
-    load_dotenv()
-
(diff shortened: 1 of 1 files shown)


training/recipes/rl_loop.py
```diff
@@ -647,14 +651,59 @@ def _loop_metrics_callback(loop_metrics: dict) -> None:
...
+                    logger.warning("Cleanup: failed to scale deployment %s: %s", cfg.deployment.deployment_id, e)
 
 
 if __name__ == "__main__":
+    from dotenv import load_dotenv
+    load_dotenv()
+    
     logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
-    main(Config(log_path="./rl_logs"))
+
```
Hecate0821: We better keep this in the example, rather then loop

PR #211, opened 2026-03-15
fix(rl): reuse policy trainer for lora grpo references

benjibc:
## Summary
- reuse the policy trainer for LoRA GRPO reference forwards by opening a base-model reference client on the same trainer instead of launching a second forward-only job
- preserve the existing explicit reference path when a reference job, base URL, or reference training shape is provided, and mirror the same reuse logic in the FrozenLake GRPO example
- add unit coverage for the manual `model_id` override path and the LoRA reference reuse flow, and extend the GRPO smoke harness so it can validate LoRA reuse with env-driven resource overrides

## Testing
- [x] `uvx ruff check training/utils/__init__.py training/utils/client.py training/recipes/rl_loop.py training/examples/frozen_lake/train_frozen_lake.py training/tests/smoke_test/test_grpo_smoke.py training/tests/unit/test_client.py training/tests/unit/test_rl_loop.py --ignore E402`
- [x] `../.venv/bin/python -m pytest tests/unit/test_client.py tests/unit/test_rl_loop.py tests/test_smoke_imports.py -q`
- [ ] `../.venv/bin/python -m pytest tests/smoke_test/test_grpo_smoke.py -q -s`

## Notes
- the remote GRPO smoke harness now accepts LoRA-specific env overrides (`FIREWORKS_SMOKE_GRPO_*`) so the reuse path can be exercised either with a fresh validated shape or with pre-created policy/deployment resources
- I attempted the remote LoRA smoke earlier, but fresh launches were blocked by validated `LORA_TRAINER` shape availability on the account and a pre-created trainer I could reuse was short-lived; the local unit coverage here is green, and the smoke harness changes are included so the live check can be retried from a compatible environment

## Diagram
```mermaid
flowchart LR
    A[GRPO config] --> B{LoRA + grpo + kl_beta != 0\nwithout explicit reference config}
    B -- yes --> C[Create policy trainer once]
    C --> D[Policy client]
    C --> E[Base-model reference client\non same trainer]
    D --> F[forward/backward + optim step]
    E --> G[reference logprobs\nwith LoRA adapters disabled]
    B -- no --> H[Existing separate reference path]
```

### DIFF
--- a/training/examples/frozen_lake/train_frozen_lake.py
+++ b/training/examples/frozen_lake/train_frozen_lake.py
@@ -24,7 +24,7 @@ import sys
 import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, cast
 
 import tinker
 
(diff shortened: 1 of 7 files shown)


Hecate0821:

PR #263, opened 2026-03-26
feat: add deployment_checkpoint_save_interval for periodic deployable checkpoints

Hecate0821:
## Summary
- Adds `deployment_checkpoint_save_interval` to `WeightSyncConfig` (default 0 = disabled)
- When enabled, periodically saves sampler-only (deployable) checkpoints during RL training
- These checkpoints can be promoted to standalone models or used to create separate deployments (e.g. for evaluation) without interfering with the training deployment's hotload delta chain

## Motivation
Users want to run evaluation on a separate deployment during training. Instead of managing a second deployment with synchronized hotloading (which complicates the delta checkpoint chain), this provides a simpler path: save periodic deployable checkpoints that users can independently deploy from.

## Test plan
- [ ] Existing unit tests pass (`pytest tests/unit/test_rl_loop.py`)
- [ ] Run RL loop with `deployment_checkpoint_save_interval=0` (default) — no behavior change
- [ ] Run RL loop with `deployment_checkpoint_save_interval=N` — verify `deploy-step-{step}` sampler checkpoints appear in `checkpoints.jsonl`

### DIFF
--- a/training/recipes/rl_loop.py
+++ b/training/recipes/rl_loop.py
@@ -766,6 +766,27 @@ def main(
                     training_shape=cfg.infra.training_shape_id,
                 )
 
+            if cfg.weight_sync.deployment_checkpoint_save_interval > 0 and step % cfg.weight_sync.deployment_checkpoint_save_interval == 0:
+                logger.info("[step %d] deployment_checkpoint_save...", step)
+                t0 = _time.time()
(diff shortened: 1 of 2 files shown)


Hecate0821:
Closing: design doesn't hold up — promote requires a running trainer job, so saving sampler-only checkpoints isn't useful. The existing dcp_save_interval + train_promote_checkpoint.py flow already covers this. Will revisit with a better design.

PR #270, opened 2026-03-27
training: bound sampler checkpoint export

benjibc:
## Summary
- honor `ReconnectableClient.save_weights_for_sampler_ext(...)` timeouts instead of blocking indefinitely
- add phase-specific timing logs for DCP save vs sampler export in `save_checkpoint(...)`
- add focused unit coverage for sampler export timeout handling and checkpoint phase logging

## Why
We reproduced multiple training runs that completed optimization steps and then stalled at `Saving final checkpoint...` before promotion. The cookbook is not the root cause of the server-side export stall, but it currently makes the failure opaque:
- `save_weights_for_sampler_ext(...)` advertises a timeout and ignores it
- `save_checkpoint(...)` does not distinguish DCP save from sampler export in logs

This patch makes the cookbook fail clearly and leaves cleaner evidence for backend debugging.

## Testing
- `training/.venv/bin/pytest training/tests/unit/test_client.py -q`
- `training/.venv/bin/pytest training/tests/unit/test_checkpoint_utils.py -q -k TestSaveCheckpoint`

### DIFF
--- a/training/tests/unit/test_checkpoint_utils.py
+++ b/training/tests/unit/test_checkpoint_utils.py
@@ -1,6 +1,7 @@
 """Unit tests for checkpoint_utils -- resume, save."""
 
 import json
+import logging
 import os
 import tempfile
 from unittest.mock import MagicMock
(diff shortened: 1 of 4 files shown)


Hecate0821:

PR #299, opened 2026-04-06
feat: LoRA self-reference RL — use policy trainer as KL reference

mayinghan:
When training with LoRA, the base model (without adapter) can serve as the KL divergence reference model.  This eliminates the need for a separate FORWARD_ONLY reference trainer job, halving reference GPU cost.

Changes:
- Add BaseReferenceClient: lightweight forward-only wrapper for base model logprobs, sharing the policy client's service session
- Add ReconnectableClient.create_base_reference(): creates a base-only model handle (base-<hex>) on the same LORA_TRAINER, with LoRA adapters disabled for forward passes
- Store _svc on ReconnectableClient so the service session can be reused for both policy and reference model handles
- New cookbook example: training/examples/rl/lora_self_ref/ demonstrating GRPO with LoRA where no separate reference training shape is needed

Depends on SDK changes:
- tinker CreateModelRequest.base_only field
- FiretitanServiceClient.create_base_training_client() method

### DIFF
--- a/training/examples/rl/lora_self_ref/train_lora_self_ref.py
+++ b/training/examples/rl/lora_self_ref/train_lora_self_ref.py
@@ -1,0 +1,688 @@
+#!/usr/bin/env python3
+"""LoRA RL training with self-reference: no separate reference trainer needed.
+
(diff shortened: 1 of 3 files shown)


Hecate0821:

PR #308, opened 2026-04-07
feat(training): add accelerator_type filtering to training shape selection

hershalb:
## Description

Add `accelerator_type` filtering to the training shape selector so jobs forced onto a specific accelerator (e.g. B200 via `UseTrainingV2`) fail fast when no matching shapes exist, instead of silently selecting the wrong accelerator's shapes and failing later with a cryptic 400.

**Companion PR:** https://github.com/fw-ai/fireworks/pull/21714 (updates `training_shape_utils.py` to thread `accelerator_type` from the job config)

### Changes

- `ShapeSelectionRequest`: new optional `accelerator_type` field (backward compatible default `None`)
- `_build_latest_validated_training_shape_filter` / `_build_compatible_training_shape_filter`: add `snapshot.accelerator_type` clause to server-side filter when set
- `_compatible_training_shape_candidates`: client-side accelerator check as belt-and-suspenders
- `_select_training_shape_candidate`: thread `accelerator_type` to both filter builders
- Split error reporting:
  - `_format_internal_shape_error`: full constraint details logged at ERROR (base_model, accelerator_type, mode, seq_len)
  - `_format_user_facing_shape_error`: generic message raised to users ("No training configuration available")
- 4 new unit tests covering server filter, client filter, mismatch error, and backward compat

## Testing

All 12 tests in `test_training_shapes.py` pass.

### DIFF
--- a/training/tests/unit/test_training_shapes.py
+++ b/training/tests/unit/test_training_shapes.py
@@ -307,7 +307,7 @@ def test_select_validated_launch_shapes_raises_when_no_validated_shape_matches()
         raise AssertionError(f"unexpected path: {path}")
 
     mgr = _FakeSelectorMgr(handler)
-    with pytest.raises(ValueError, match="No validated training shape matched"):
+    with pytest.raises(ValueError, match="No training configuration is available"):
         select_validated_launch_shapes(
             mgr,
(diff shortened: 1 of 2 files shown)


Hecate0821:

PR #330, opened 2026-04-14
fix(dpo): construct reference client with lora_rank=0

Hecate0821:
## Summary

Companion to #326. That PR fixed the reference *trainer* creation to use `lora_rank=0` (the ref trainer only runs base-model forward passes, no adapters), but the matching `ReconnectableClient` was still constructed with `cfg.lora_rank`, so the client's `create_model` call requests LoRA mode on a trainer initialized with `--lora-rank=0`:

```
tinker.RequestFailedError: create_model requested LoRA mode, but trainer
was initialized without LoRA config. Check RLOR startup args include --lora-rank.
```

Reproduced on [run 24419618895](https://github.com/fw-ai/fireworks/actions/runs/24419618895) for kimi-k2p5-text-only-256k-lora.

## Fix

One line — pin the reference `ReconnectableClient`'s `lora_rank` to `0` so client and trainer agree.

## Test plan

- [ ] Re-run the kimi-k2p5-text-only-256k-lora single-shape CI; should now reach DPO step 1

### DIFF
--- a/training/recipes/dpo_loop.py
+++ b/training/recipes/dpo_loop.py
@@ -547,7 +547,8 @@ def main(
         reference_job_id = reference_ep.job_id
 
         policy = ReconnectableClient(rlor_mgr, policy_ep.job_id, cfg.base_model, cfg.lora_rank)
-        reference = ReconnectableClient(rlor_mgr, reference_ep.job_id, cfg.base_model, cfg.lora_rank)
+        # Match the ref trainer's lora_rank=0 (set above).
+        reference = ReconnectableClient(rlor_mgr, reference_ep.job_id, cfg.base_model, 0)
         if hasattr(policy, "close"):
             stack.callback(policy.close)
         if hasattr(reference, "close"):


training/recipes/dpo_loop.py
```diff
@@ -547,7 +547,8 @@ def _on_trainer_status(msg: str) -> None:
         reference_job_id = reference_ep.job_id
 
         policy = ReconnectableClient(rlor_mgr, policy_ep.job_id, cfg.base_model, cfg.lora_rank)
-        reference = ReconnectableClient(rlor_mgr, reference_ep.job_id, cfg.base_model, cfg.lora_rank)
+        # Match the ref trainer's lora_rank=0 (set above).
+        reference = ReconnectableClient(rlor_mgr, reference_ep.job_id, cfg.base_model, 0)
```
morgendave, line 551: Does it work with lora itself?

Hecate0821: We still need this to align https://github.com/fw-ai/cookbook/pull/330/changes#diff-46b01eada427263be154e80119718ce86b6a1d9f13f84c726af27b1b52123d4eL520

morgendave (approved):

PR #335, opened 2026-04-15
fix: fail fast when training data doesn't fill a single batch

xiaoyifan:
Previously, if filtered examples existed but their total token count didn't fill a single batch, sft_loop would report 0 batches/epoch, run 0 optimizer steps, and mark the job as COMPLETED. This is misleading — the customer pays for GPU time but gets no training.

Now raise RuntimeError with an actionable message when total_batches_per_epoch == 0, telling the user to provide more data, reduce batch_size, or increase max_seq_len.

Observed on job rft--pyroworks-dev-b9dr5fi1: 82/100 examples filtered (len > 32768), 18 remaining examples didn't fill a batch, resulted in 'Training complete: 0 optimizer steps' with COMPLETED status.

### DIFF
--- a/training/recipes/sft_loop.py
+++ b/training/recipes/sft_loop.py
@@ -455,6 +455,12 @@ def main(
             total_batches_per_epoch,
             cfg.epochs,
         )
+        if total_batches_per_epoch == 0:
+            raise RuntimeError(
+                f"No training batches: {len(training_data)} examples remaining after "
+                f"filtering do not fill a single batch (batch_size={cfg.batch_size}). "
+                f"Provide more data, reduce batch_size, or increase max_seq_len."
+            )
         if eval_data:
             logger.info("Eval dataset: %d examples (eval after each epoch)", len(eval_data))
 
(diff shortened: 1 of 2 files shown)


Hecate0821 (approved):

training/recipes/sft_loop.py
```diff
@@ -455,6 +455,12 @@ def _map_fn(row: dict) -> tinker.Datum | None:
             total_batches_per_epoch,
             cfg.epochs,
         )
+        if total_batches_per_epoch == 0:
```
websterbei, line 458: this is because PP?

training/recipes/sft_loop.py
```diff
@@ -455,6 +455,12 @@ def _map_fn(row: dict) -> tinker.Datum | None:
...
+            raise RuntimeError(
+                f"No training batches: {len(training_data)} examples remaining after "
+                f"filtering do not fill a single batch (batch_size={cfg.batch_size}). "
```
websterbei, line 461: Is there a minimum number of samples needed? does 1 sample not work?

PR #343, opened 2026-04-16
feat(rl): skip separate reference trainer for LoRA GRPO

morgendave:
## Summary

When running LoRA GRPO with KL regularisation (`kl_beta > 0`), reference logprobs can come from the **same trainer** via a base-only model handle (adapters disabled). This eliminates provisioning a second forward-only trainer job, halving GPU cost for LoRA RL.

### What changed

**`training/recipes/rl_loop.py`**
- Reference resolution uses direct config checks (`cfg.kl_beta`, `cfg.lora_rank`, `cfg.infra.ref_training_shape_id`) — no intermediate enums or booleans
- `_make_client()` helper deduplicates trainer client creation (incorporates `cfg.step_timeout`)
- LoRA + `kl_beta > 0` without explicit ref shape: calls `policy.create_base_reference()` — no second job
- Explicit `ref_training_shape_id` or full-param KL: provisions a separate forward-only trainer
- `kl_beta=0`: no reference at all

**`training/utils/client.py`**
- `ReconnectableClient.create_base_reference()` returns a new client sharing the same trainer with a `base-<hex>` model handle
- Uses SDK's `FiretitanServiceClient.create_base_training_client()` (`fireworks-ai>=1.0.0a60`) — no raw HTTP workaround
- Added `base_only` parameter to `ReconnectableClient`

**`training/examples/rl/deepmath/train_deepmath.py`**
- Added `--lora-rank` CLI argument so customers can run LoRA GRPO directly

**`training/tests/unit/test_rl_loop.py`**
- Added `test_lora_shared_reference_creates_single_trainer` — validates only one trainer job is provisioned and `create_base_reference()` is called for LoRA + KL

### Customer usage

```bash
python -m examples.rl.deepmath.train_deepmath \
    --base-model accounts/fireworks/models/qwen3p5-32b \
    --lora-rank 64 \
    --dataset my_math_prompts.jsonl
```

The cookbook auto-selects the policy training shape. When `lora_rank > 0` and no explicit `ref_training_shape_id` is set, reference logprobs come from the policy trainer via a base-only handle.

### Before / After

| | Before | After |
|---|---|---|
| LoRA GRPO trainers | 2 (policy + reference) | 1 (policy only) |
| GPU cost | 2× | 1× |
| Control flow | scattered booleans | direct config checks in if/elif/else |

## Test plan

- [ ] LoRA + `kl_beta > 0`, no `ref_training_shape_id` → single trainer, `create_base_reference()` called
- [ ] Full-param + `kl_beta > 0` → separate reference trainer provisioned
- [ ] Explicit `ref_training_shape_id` → separate reference trainer provisioned
- [ ] `kl_beta=0` → no reference client created
- [ ] `--lora-rank` flag works in `train_deepmath.py`
- [ ] Unit test passes: `test_lora_shared_reference_creates_single_trainer`

### DIFF
--- a/training/recipes/rl_loop.py
+++ b/training/recipes/rl_loop.py
@@ -91,6 +91,7 @@ from training.utils.rl.router_replay import build_r3_routing_matrices
 
 logger = logging.getLogger(__name__)
 
+
 # ---------------------------------------------------------------------------
 # Config
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


benjibc (approved):

training/utils/client.py
```diff
@@ -242,3 +282,96 @@ def _normalize_grad_accumulation_normalization(
         raise ValueError(
             f"Unknown grad_accumulation_normalization '{value}'. Expected one of: {valid}"
         ) from exc
+
+
+# ---------------------------------------------------------------------------
+# Base-only client (raw HTTP, no SDK dependency on base_only flag)
```
Hecate0821: We should put this into SDK

training/recipes/rl_loop.py
```diff
@@ -360,8 +354,12 @@ def _signal_handler(signum, frame):
 
     # -- Resolve reference training shape --------------------------------------
     reference_needed = cfg.kl_beta > 0 or cfg.infra.ref_training_shape_id is not None
+    # LoRA optimisation: the policy trainer already has the frozen base model,
+    # so we can obtain reference logprobs by creating a base-only handle on
+    # the same trainer — no second GPU allocation needed.
+    _lora_shares_trainer = cfg.lora_rank > 0 and not cfg.infra.ref_training_shape_id
     ref_profile = None
-    if reference_needed:
+    if reference_needed and not _lora_shares_trainer:
```
Hecate0821: Can we make the control flow cleaner? We have 5 bool for 3 states now
```
  class RefMode(Enum):                                                                                                                                         
      NONE = auto()                                                                                                                                            
      SHARED = auto()      # base-only handle, same trainer
      SEPARATE = auto()    # second trainer job                                                                                                                
                                                   
```

Hecate0821 (commented):
And can you update the skill about thjis

training/recipes/rl_loop.py
Hecate0821: Or we can eliminate the isolated ref model path

training/utils/client.py
morgendave: Let me update. Thought AGI handled it correctly

training/recipes/rl_loop.py
morgendave: ref is shared between full and RL. Let me check if we can just upgrade

PR #362, opened 2026-04-21
feat(training/renderer): add glm5 renderer for GLM-5.1

mayinghan:
Adds a new `glm5` renderer covering ZhipuAI's GLM-5.1 instruct models.

Chat template layout:

    [gMASK]<sop><|user|>
    {user_content}<|assistant|>
    <think></think>
    {assistant_content}<|endoftext|>

Key details:
- BOS `[gMASK]<sop>` emitted once via `_bos_tokens`.
- Role tags `<|system|>`, `<|user|>`, `<|assistant|>`, `<|observation|>`.
- Historical assistant turns collapse thinking to `<think></think>` (controlled by `strip_thinking_from_history`, default True).
- Appends `<|endoftext|>` to trained assistant turns so the model learns to stop — matches the convention already used by the Qwen3, Nemotron, and MiniMax M2 renderers in this repo.

Validated end-to-end via SFT memorization training on `accounts/fireworks/models/glm-5p1`. Auto-selection added to `resolve_renderer_name` for `glm-5p1`, `glm-5.1`, `glm5`.


training/renderer/glm5.py
```diff
@@ -0,0 +1,219 @@
...
+            if not isinstance(part, Mapping):
+                continue
+            if part.get("type") == "thinking" and isinstance(part.get("thinking"), str):
+                reasoning_parts.append(part["thinking"])
+            elif part.get("type") == "text" and isinstance(part.get("text"), str):
+                text_parts.append(part["text"])
+    return "".join(reasoning_parts), "".join(text_parts)
+
+
+class GLM5Renderer(Renderer):
```
lequytra, line 98: Have you validate the renderer w/ HF's chat template in thorough unit test cases? I found AGI usually misses a lot of chat template rules that needs to be test pretty thoroughly. You can use gemma, nemotron as examples

benjibc (approved):

training/renderer/glm5.py
mayinghan: super legit, caught a multiturn issue in the test

PR #363, opened 2026-04-21
fix(skill-docs): align server-side validation error table with merged CP

Hecate0821:
## Summary

- Drop the phantom \`hotload bucket mismatch: deployment Y has hot_load_bucket_url=P but a trainer with hot_load_deployment_id=X would write to Q\` row from the server-side validation table in both \`skills/dev/references/rl/hotload.md\` and \`skills/dev/SKILL.md\`. No merged CP code emits that string (verified via grep across fw-ai/fireworks and the merged fw-ai/fireworks#22823 diff).
- Add the real create-vs-update variant the CP does emit: \`hot_load_bucket_url %q conflicts with hot_load_trainer_job %s; update hot_load_trainer_job instead, or clear it first\` on \`UpdateDeployment\`.

The same phantom row currently lives in the docs PR fw-ai/docs#886; flagging separately on that PR.

## Test plan

- [x] \`grep -c "hotload bucket mismatch: deployment Y" skills/dev/references/rl/hotload.md\` returns 0
- [x] \`grep -c "update hot_load_trainer_job instead" skills/dev/references/rl/hotload.md\` returns 1
- [ ] Reviewer cross-checks error list against fw-ai/fireworks#22823 (\`control_plane/pkg/gateway_server/deployment.go\` + \`rlor_trainer_job.go\`)


Hecate0821:
Folded into #351 — that branch was rebased onto main (the original narrative was subsumed by #356) and this PR's single commit was cherry-picked on top. #351 now carries the same fix with a cleaner base.

PR #371, opened 2026-04-22
fix(sft): stream-render dataset to fix orchestrator OOM on large/long-context jobs

xiaoyifan:
## Summary

Fixes orchestrator (cookbook runner) OOMs during SFT dataset rendering on large / long-context datasets by switching from eager full-dataset tokenization to a streaming PyTorch `DataLoader` pipeline.

Instead of materializing `List[tinker.Datum]` in the orchestrator, this PR scans JSONL offsets once, renders rows on demand inside spawn workers, and feeds batches directly into training. Peak RAM is now bounded by roughly `num_workers × per-worker render footprint`, not by the total rendered dataset size.

Validated end-to-end on job `r62fmz8e` (110,919 examples × 256K context, `qwen3p5-397b-a17b`). This PR also rebases onto current `main` and incorporates the follow-up review fixes for deterministic resume + zero-valid-data handling.

Pairs with **fw-ai/fireworks#23033** (orchestrator memory/disk allocation update for the streaming renderer) and the deeper investigation in `docs/engineering/sft-v2-orchestrator-oom-debug.md`.

## Architecture

```mermaid
flowchart LR
    A[JSONL on disk / GCS-localized file] --> B[_scan_jsonl_offsets]
    B --> C[JsonlRenderDataset]
    C -->|getitem -> json.loads + render_fn| D[PyTorch DataLoader]
    D -->|spawn workers| W1[worker 1]
    D --> W2[worker 2]
    D --> W3[worker 3]
    D --> W4[worker 4]
    W1 --> T[trainer.forward_backward]
    W2 --> T
    W3 --> T
    W4 --> T
    E[eval carve-out / explicit eval] -->|bounded eager render| F[eval_data list]
```

### `training/utils/streaming.py`

- Add `JsonlRenderDataset`: map-style dataset that scans JSONL byte offsets once, then lazily `seek + readline + json.loads + render_fn` per row.
- Add `make_render_dataloader`: wraps PyTorch `DataLoader` with our spawn / persistent-worker / drop-`None` defaults.
- Add explicit `generator=` plumbing so callers can control shuffle determinism.

### `training/recipes/sft_loop.py`

- Replace eager `raw_data.append(...)` + full in-memory rendered dataset with streaming `JsonlRenderDataset` + `DataLoader`.
- Keep eval rendering bounded: explicit eval datasets are still rendered eagerly; auto carve-out still only materializes the small eval slice.
- Restore deterministic epoch ordering by seeding the loader generator as `cfg.seed + epoch`.
- On resume, replay epoch-0 with the same seed, then skip already-consumed batches via `islice(...)`.
- Raise `RuntimeError("No valid training examples after tokenization")` if epoch 0 renders zero valid rows, instead of silently completing with 0 optimizer steps.
- Log filtered-row count after epoch 0.

### `training/utils/infra.py`

- Add a tiny SDK-compat shim so `TrainerJobConfig(skip_validations=...)` is only passed when the installed SDK supports that kwarg. This keeps the SFT loop unit tests runnable against older local SDK builds while preserving current behavior in CI / newer SDKs.

### Tests

- `test_streaming.py`: cover `generator=`-based deterministic shuffle.
- `test_sft_loop.py`: cover deterministic resume order and the zero-valid-data failure path; update the per-batch optimizer-step test to stop assuming input order when shuffle is enabled.

## Why streaming?

Earlier iterations failed because eager rendering scaled orchestrator memory with the fully tokenized dataset. For long-context / large-example-count jobs, that either OOMed the pod or pushed requested memory above what the node could schedule. Streaming sidesteps that by making render memory independent of total dataset size.

## Type of Change

- [x] Bug fix
- [x] Refactoring

## Testing

- [x] End-to-end validated on job `r62fmz8e`
- [x] `pytest -q training/tests/unit/test_streaming.py`
- [x] `pytest -q training/tests/unit/test_sft_loop.py`
- [x] `pytest -q training/tests/unit/test_streaming.py training/tests/unit/test_sft_loop.py`

## Follow-ups (out of scope)

- DPO / ORPO migration to the same streaming render path
- Further cleanup of shared render-worker init boilerplate (handled separately in follow-up PRs)

### DIFF
--- a/training/recipes/sft_loop.py
+++ b/training/recipes/sft_loop.py
@@ -14,9 +14,11 @@ Usage:
 
 from __future__ import annotations
 
+import functools
 import os
 import signal
 import logging
+import tempfile
 from contextlib import ExitStack
 from typing import Any, Dict, List
 from dataclasses import field, dataclass
@@ -24,7 +26,6 @@ from dataclasses import field, dataclass
 import torch
 import tinker
 
-import json
 import datasets as hf_datasets
 import transformers
 from dotenv import load_dotenv
@@ -33,13 +34,20 @@ from fireworks.training.sdk import TrainerJobManager
 from tinker_cookbook.supervised.data import SupervisedDatasetFromHFDataset
 from training.utils import (
     DEFAULT_ADAM,
+    DEFAULT_RENDER_CHUNKSIZE,
+    DEFAULT_RENDER_WORKERS,
+    DiskBackedDatumStore,
     InfraConfig,
+    MemTracer,
     ResourceCleanup,
     RunnerConfig,
     RunnerIO,
     RunStatus,
     WandBConfig,
     ReconnectableClient,
+    count_jsonl_rows,
+    iter_jsonl_rows,
+    stream_render_to_store,
     wandb_log,
     setup_wandb,
     wandb_finish,
@@ -65,6 +73,52 @@ from training.utils.timer import timer, flush_timing
 
 load_dotenv()
 logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Per-worker render state (multiprocessing) -- module-level so the spawn
+# pool initializer can populate it before each worker starts consuming rows.
+# ---------------------------------------------------------------------------
+
+_worker_renderer = None
+_worker_train_on_what = None
+_worker_max_seq_len = None
+
+
+def _init_render_worker(tokenizer_model, renderer_name, train_on_what_str, max_seq_len):
+    """Build a renderer per worker (avoids pickling the tokenizer)."""
+    global _worker_renderer, _worker_train_on_what, _worker_max_seq_len
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+        tokenizer_model, trust_remote_code=True,
+    )
+    _worker_renderer = build_renderer(tokenizer, tokenizer_model, renderer_name)
+    _worker_train_on_what = parse_train_on_what(train_on_what_str)
+    _worker_max_seq_len = max_seq_len
+
+
+def _render_messages(
+    row: dict, *, renderer, train_on_what, max_seq_len: int,
+) -> tinker.Datum | None:
+    """Render a chat row to a Datum, filtering empty / out-of-range sequences."""
+    messages = row.get("messages", [])
+    if not messages:
+        return None
+    rendered = render_messages_to_datum(
+        messages, renderer=renderer, train_on_what=train_on_what,
+    )
+    if len(rendered.token_ids) > max_seq_len or len(rendered.token_ids) < 2:
+        return None
+    return rendered.datum
+
+
+def _render_one_worker(row: dict) -> tinker.Datum | None:
+    """Render a single chat example in a worker process."""
+    return _render_messages(
+        row,
+        renderer=_worker_renderer,
+        train_on_what=_worker_train_on_what,
+        max_seq_len=_worker_max_seq_len,
+    )
+
 
 # ---------------------------------------------------------------------------
 # Eval auto carve-out defaults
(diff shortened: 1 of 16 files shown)


mayinghan (approved):

Hecate0821:
Can it handle resume? looks like it is shuffled, so the resume may not semantically correct. And seems like resume still render skipped rows.

training/recipes/sft_loop.py
```diff
@@ -595,10 +610,16 @@ def _run_train_step(
...
+            # On resume, skip the batches already processed in epoch 0.
+            # Workers still render skipped batches (wasted but bounded by
+            # num_workers * prefetch_factor); resume is rare enough that
+            # this is acceptable.
+            batch_iter = iter(loader)
+            if epoch == 0 and start_batch > 0:
+                batch_iter = islice(batch_iter, start_batch, None)
+            for batch in batch_iter:
+                if not batch:
+                    continue  # entire batch was filtered (None render); skip
```
renfeichen-fw, line 718: **Silent zero-steps regression.** Combined with the line-159 check (which only catches an empty JSONL, not post-render emptiness), a dataset where every row renders to `None` (too long, bad format, etc.) now "succeeds" with 0 optimizer steps and no output model — while still reporting `RunStatus.COMPLETED` and billing the customer.

Please track a counter of non-None rendered rows and raise if it is 0 at end of epoch (or at job end). Pre-PR this was caught eagerly in `_map_fn`.

training/recipes/sft_loop.py
```diff
@@ -375,126 +463,53 @@ def _on_trainer_status(msg: str) -> None:
...
-            effective_batch_size = len(training_data)
 
-        sft_dataset = SupervisedDatasetFromHFDataset(
-            hf_datasets.Dataset.from_dict({"datum_idx": list(range(len(training_data)))}),
+        loader = make_render_dataloader(
+            training_dataset,
             batch_size=effective_batch_size,
-            map_fn=lambda row: training_data[row["datum_idx"]],
+            num_workers=num_workers,
+            shuffle=True,
```
renfeichen-fw, line 562: **Lost reproducibility + non-deterministic resume.** `shuffle=True` uses `torch.default_generator`, which is unseeded here. The pre-PR path (`SupervisedDatasetFromHFDataset`) called `hf_dataset.shuffle(seed=seed)` inside `set_epoch`, so the same `(dataset, seed, epoch)` gave the same order.

Now: (a) two runs with the same config see different batch orders, and (b) resumes replay a *different* random sequence than the original run. Pass an explicit seeded `torch.Generator` (derived from `cfg.seed`) via the `generator=` kwarg.

training/recipes/sft_loop.py
```diff
@@ -595,10 +610,16 @@ def _run_train_step(
...
-            epoch_start = start_batch if epoch == 0 else 0
-            for i_batch in range(epoch_start, total_batches_per_epoch):
-                batch = sft_dataset.get_batch(i_batch)
+            # On resume, skip the batches already processed in epoch 0.
+            # Workers still render skipped batches (wasted but bounded by
+            # num_workers * prefetch_factor); resume is rare enough that
+            # this is acceptable.
+            batch_iter = iter(loader)
+            if epoch == 0 and start_batch > 0:
+                batch_iter = islice(batch_iter, start_batch, None)
```
renfeichen-fw: `iter(loader) + islice(..., start_batch, None)` is not resume-safe: workers spin up, prefetch ahead, then the first `start_batch` batches are discarded — but prefetch draws are not rolled back, so post-resume order diverges from a non-resumed run. If we must keep this, at minimum document in the comment that resumed runs see a different data sequence than fresh runs. Better: use a seeded `DistributedSampler` / `RandomSampler` and skip by sample-index rather than batch-iter.

renfeichen-fw (changes requested):
Two blocking concerns — silent zero-steps and lost resume determinism. See inline.

training/recipes/sft_loop.py
```diff
@@ -375,126 +463,53 @@ def _on_trainer_status(msg: str) -> None:
...
-        total_batches_per_epoch = len(sft_dataset)
+        # Pre-filter upper bound; filtered rows make actual batches
+        # smaller but never larger.
+        total_batches_per_epoch = (training_count + effective_batch_size - 1) // effective_batch_size
         logger.info(
-            "Dataset: %d examples, %d batches/epoch, %d epochs",
-            len(training_data),
-            total_batches_per_epoch,
-            cfg.epochs,
+            "Dataset: %d examples from %s (renderer=%s, train_on_what=%s,"
```
renfeichen-fw, line 570: **Rendering progress + filter count dropped.** Pre-PR the loop called `runner.report_rendering_progress(i, total)` at ~5 % intervals and logged `"Seq-length filter: N/M examples filtered"`. Both are gone.

For the 110 K × 256 K dataset that motivated this PR, the render phase alone takes ~30 min. During that window the orchestrator now receives zero progress updates, which could trigger hung-job alerts and confuse customers tracking long-running jobs.

Consider: (a) a periodic callback inside the DataLoader iteration (e.g. log after every ~N batches), and (b) accumulating the filtered count in `_render_one_worker` (e.g. via a module-level counter or by returning a sentinel the consumer can tally) and logging it after the first epoch.

training/tests/unit/test_sft_loop.py
```diff
@@ -149,6 +149,7 @@ def __init__(self, *args, **kwargs):
         dataset=str(dataset_path),
         tokenizer_model="Qwen/Qwen3-4B",
         max_seq_len=32,
+        render_workers=1,
     )
 
     with pytest.raises(RuntimeError, match="No valid training examples"):
```
renfeichen-fw, line 167: **Existing test will break.** This assertion expects `RuntimeError("No valid training examples")` but the new code only raises `RuntimeError("No examples found in {path}")` — and only for empty files. When the file has rows but every row renders to `None` (as this test's dataset does: all empty messages), no error is raised and `main()` completes with 0 steps.

This is the test-side manifestation of the silent-zero-steps regression flagged on `sft_loop.py:622`. Once that fix lands, update this assertion to match the new error message.

training/recipes/sft_loop.py
```diff
@@ -167,6 +249,12 @@ class Config:
     max_eval_seqs: int = DEFAULT_MAX_EVAL_SEQS
     """Max number of eval sequences for carve-out."""
 
+    render_workers: int | None = None
```
renfeichen-fw, line 285: **Merge drift vs current `origin/main` — this PR's base is stale.** The PR's base SHA `eae3e6f2` is not an ancestor of today's `main` (`5b700ca`). Current `main` adds three `Config` fields that do not exist on this branch: `adam_beta2` (:146), `weight_decay` (:151), `warmup_steps` (:154), plus their wiring into `adam_kwargs` and the warmup LR schedule at :556-565.

If this PR merges as-is (or gets force-rebased wrong), those fields silently disappear. Impact: customer configs that set `adam_beta2=`/`weight_decay=`/`warmup_steps=` break at `Config(...)` construction, and jobs relying on `warmup_steps > 0` train with constant LR from step 0.

Please rebase onto current `main` and re-add these fields before merge.

training/recipes/sft_loop.py
```diff
@@ -375,126 +463,53 @@ def _on_trainer_status(msg: str) -> None:
...
-        renderer = build_renderer(tokenizer, cfg.tokenizer_model, cfg.renderer_name)
-        train_on_what = parse_train_on_what(cfg.train_on_what)
-        logger.info(
-            "Using renderer=%s train_on_what=%s",
-            resolve_renderer_name(cfg.tokenizer_model, cfg.renderer_name),
-            train_on_what.value,
+        # Render JSONL rows on the fly inside DataLoader workers so peak
+        # RAM is O(num_workers * per_worker_render_footprint) instead of
+        # O(dataset_size). See docs/engineering/sft-v2-orchestrator-oom-debug.md.
+        num_workers = (
```
renfeichen-fw, line 532: **Manual infra path also missing from this branch's base.** Current `origin/main` has an `if not training_shape_id and (accelerator_type or node_count or custom_image_tag or extra_args): manual path` branch right above this block — lets callers launch with direct overrides when no validated shape exists. It is absent here because this branch is behind main.

Same concern as the Config fields: after rebase-and-merge, the manual-infra branch disappears and any caller using `custom_image_tag`/`accelerator_type`/`node_count`/`extra_args` without `training_shape_id` hard-fails at shape lookup.

Please rebase onto current `main` and confirm both the manual-infra branch and the Config fields survive the merge.

xiaoyifan:
Pushed follow-up fixes on top of the rebase to current `main` (`b78c7e9`).

Addressed from review:
- deterministic epoch ordering / resume: the SFT loader now uses an explicit `torch.Generator`, reseeded per epoch as `cfg.seed + epoch`, so a resumed run replays the same epoch-0 shuffle before skipping already-consumed batches
- zero-valid-data guard: epoch 0 now raises `RuntimeError(\"No valid training examples after tokenization\")` instead of completing with 0 optimizer steps
- filtered-row visibility: after epoch 0 we now log `Seq-length / format filter: N/M raw rows filtered`
- stale-base concerns: rebased onto current `main`, so the newer `Config` fields + manual infra path are preserved

I also refreshed the PR body to match the current DataLoader-based design (the older disk-store / MemTracer description was stale).

One nuance on the render-progress comment: training rows are no longer pre-rendered in a separate long-running phase; they are streamed on demand during training, so there is no standalone "render 110k rows, then train" stage to report the old `runner.report_rendering_progress(...)` against. The new code keeps the filter-count visibility, while training progress continues to come from the normal step-level status updates.

xiaoyifan:
Latest update on top of `3f43890`:

- merged the branch forward to current `main` (including `#388`) so the PR is no longer conflicting
- kept the streaming/DataLoader SFT path, but also carried over the newer SFT semantics from `main` that are still relevant here:
  - `seed` now drives both deterministic resume ordering and eval auto-carveout membership
  - auto-carveout now picks a seeded random eval subset instead of always taking the first `N` raw rows
- refreshed the SFT tests accordingly and kept the new mainline DPO/ORPO/loss tests in the branch
- PR `Training CI` is green on the pull_request run: [24880998724](https://github.com/fw-ai/cookbook/actions/runs/24880998724)

This should be ready for another review pass.

training/recipes/sft_loop.py
xiaoyifan: Addressed on the latest head.

We now track `epoch_valid_examples` during epoch 0 and raise `RuntimeError(\"No valid training examples after tokenization\")` if nothing survives rendering, instead of silently completing with 0 optimizer steps. The regression test in `test_sft_loop.py` now covers the all-filtered case as well.

training/recipes/sft_loop.py
xiaoyifan: Fixed on the latest head.

The SFT loader now passes an explicit `torch.Generator` into `make_render_dataloader(...)` and reseeds it per epoch as `cfg.seed + epoch`, so fresh runs and resumed runs replay the same epoch-0 shuffle before skipping already-consumed batches. `test_main_resume_preserves_epoch_zero_batch_order` locks that in.

training/recipes/sft_loop.py
xiaoyifan: This one is stale relative to the current code.

We still skip by batch on resume, but the loader is now explicitly reseeded before `iter(loader)`, and `test_main_resume_preserves_epoch_zero_batch_order` covers the intended invariant for the current implementation. If we later want true sample-index resume semantics we can revisit that as a follow-up, but I don't think it blocks this PR in its current form.

training/recipes/sft_loop.py
xiaoyifan: I restored the filtered-row logging (`Seq-length / format filter: N/M raw rows filtered`) on epoch 0.

For the progress part: with the current DataLoader design there is no longer a standalone "render the whole dataset, then start training" phase. Rows are rendered on demand inside the training loop, and the orchestrator status is updated by the normal per-step training progress path. Because of that, I did not add a second render-progress status stream on top of the step-level training status.

training/tests/unit/test_sft_loop.py
xiaoyifan: Fixed.

The test now exercises the all-filtered path against the current behavior and still expects the runtime failure instead of silently completing. That regression is covered together with the epoch-0 zero-valid-data guard in the latest head.

training/recipes/sft_loop.py
xiaoyifan: Addressed.

The branch was first rebased/merged forward to current `main`, and then merged forward again after `#388` landed, so the current PR head includes the newer `Config` fields and their warmup/optimizer wiring from `main`.

training/recipes/sft_loop.py
xiaoyifan: Addressed as part of the forward merge to current `main`.

The latest PR head keeps the manual infra path from `main` in addition to the streaming SFT changes, so callers using direct overrides without a validated shape still follow the manual launch branch.

training/recipes/sft_loop.py
```diff
@@ -705,13 +689,31 @@ def _run_train_step(
...
         for epoch in range(cfg.epochs):
-            sft_dataset.set_epoch(epoch)
-            epoch_start = start_batch if epoch == 0 else 0
-            for i_batch in range(epoch_start, total_batches_per_epoch):
-                batch = sft_dataset.get_batch(i_batch)
+            loader_generator.manual_seed(cfg.seed + epoch)
+            batch_iter = iter(loader)
+            epoch_start_batch = start_batch if epoch == 0 else 0
+            if epoch_start_batch > 0:
+                batch_iter = islice(batch_iter, epoch_start_batch, None)
```
renfeichen-fw, line 708: Resume is still unsafe when rendering filters rows. Repro: `batch_size=2`, fixed seed, first raw batch has one row rendering to `None`; after one step the checkpoint records `data_consumed == 1`, so resume computes `1 // 2 == 0` and replays that already-trained example. Please add this unit test first, then fix the cursor semantics.

training/recipes/sft_loop.py
```diff
@@ -447,149 +526,55 @@ def _on_trainer_status(msg: str) -> None:
...
-            map_fn=lambda row: training_data[row["datum_idx"]],
+            num_workers=num_workers,
+            shuffle=True,
+            generator=loader_generator,
+            worker_init_fn=worker_init_fn,
         )
-        total_batches_per_epoch = len(sft_dataset)
+        # Pre-filter upper bound; filtered rows make actual batches
+        # smaller but never larger.
+        total_batches_per_epoch = (training_count + effective_batch_size - 1) // effective_batch_size
```
renfeichen-fw, line 568: This is a raw-row upper bound, but completed progress uses optimizer steps from non-empty post-filter batches. Repro with many rows rendering to `None`: the job can complete with `step < total_steps_estimate`, so COMPLETED still reports less than 100%. Please add a unit test for that first, then fix the reported total/progress behavior.

renfeichen-fw (commented):
Two additional streaming SFT edge cases to cover with regression tests first, then fix.

renfeichen-fw (approved):

training/recipes/sft_loop.py
xiaoyifan: Fixed on the latest head.

The final runner status now reports `total_steps=step` on `RunStatus.COMPLETED`, so the orchestrator converges to 100% even when filtering drops some raw batches and the training run executes fewer optimizer steps than the raw-row upper bound. I also added `test_completed_status_reports_actual_steps_when_filtering_drops_raw_batches` to lock that behavior in.

training/recipes/sft_loop.py
xiaoyifan: Fixed on the latest head.

SFT checkpoints now persist a separate `raw_rows_consumed` cursor and resume uses that raw-row cursor (falling back to `data_consumed` for older checkpoints) to derive the epoch/batch offset. That keeps resume aligned to raw DataLoader batches even when filtering shrinks a step below `batch_size`, so we no longer replay the already-trained example from the first partially-filtered batch. `test_main_resume_uses_raw_row_cursor_when_filtering_shrinks_batches` covers the exact repro from this thread.

PR #372, opened 2026-04-22
test(infra_setup): detect parallelism via paired measurement

Hecate0821:
## Summary

Rewrite `test_parallel_wait_timing` so it isolates the *wait sleep time* from non-sleep fixture/CI overhead, then checks parallelism on the isolated delta.

## Why

The existing assertion was `elapsed < 1.5 × SLEEP_S` (0.225s for `SLEEP_S=0.15s`). GHA runners add ~200ms of overhead (fixture, coverage instrumentation, `ThreadPoolExecutor` setup) on top of the sleeps, so elapsed lands ~0.36s whether the two waits actually run in parallel or not. Reruns don't help.

Observed on PR #347 CI (run 24757323518):
```
elapsed=0.381s ≥ 0.225s
elapsed=0.361s ≥ 0.225s
```

## The right fix (not just bumping SLEEP_S)

Simply increasing `SLEEP_S` would make the threshold survive CI overhead but *masks a potential regression* that makes the waits serial — the test would pass even if `_wait_for_pending_resources` accidentally went back to a sequential loop.

Paired measurement:

```python
baseline   = setup_infra(sleep=0s)    # captures all non-sleep overhead
with_sleep = setup_infra(sleep=X)     # adds actual wait time
delta      = with_sleep - baseline
assert delta < 1.5 * X                # parallel ≈ 1×X, serial ≈ 2×X
```

Fixture + coverage overhead cancels, so the 1.5× margin is a direct parallel-vs-serial discriminator. A warm-up run before the two timed runs ensures both see an equally warm interpreter.

## Local verification

```
pytest training/tests/unit/test_infra_setup.py::test_parallel_wait_timing -v
  PASSED [100%] in 2.70s

pytest training/tests/unit/test_infra_setup.py::test_parallel_wait_timing --cov=training --cov-branch -q
  1 passed in 2.29s
```

## Test plan

- [x] Passes locally, with and without coverage
- [ ] Passes on GHA (the runner class that was failing before)

### DIFF
--- a/training/tests/unit/test_infra_setup.py
+++ b/training/tests/unit/test_infra_setup.py
@@ -521,16 +521,26 @@ def test_parallel_request_phase_precedes_wait_phase(monkeypatch):
 
 
 def test_parallel_wait_timing(monkeypatch):
-    """Both trainer waits run in parallel; total wall time ≈ max(N), not sum(N)."""
+    """Both trainer waits run in parallel; elapsed grows by ~SLEEP_S, not 2×.
+
+    Uses a paired measurement: one run with ``sleep_s > 0`` and one with
+    ``sleep_s = 0`` sharing the exact same setup cost. The difference isolates
+    the wall time consumed by the two ``wait_trainer_job`` sleeps, so fixture
+    and coverage-instrumentation overhead (which can add 200ms+ on CI) does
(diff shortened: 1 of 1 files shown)


Hecate0821:
Superseded — folded the fix directly into PR #347 as commit 8ee35d5. Keeping a single PR to unblock #347 without an extra coordination step.

PR #373, opened 2026-04-22
refactor(dpo): stream via DataLoader + AppendOnlyPickleLog ref cache + LoRA ref trainer fix

xiaoyifan:
## Summary

Two DPO fixes required to run DPO V2 end-to-end on long-context / multi-epoch / LoRA jobs:

1. **Streaming dataset render via DataLoader + AppendOnlyPickleLog ref cache** — mirrors the SFT v2 rewrite (#371). Eliminates orchestrator RAM blowup on large / long-context multi-epoch runs.
2. **Propagate `lora_rank` to the reference trainer + client** — unblocks LoRA DPO V2, where the gateway's auto-selected shape is `LORA_TRAINER` mode but the cookbook was requesting the ref trainer with hardcoded `lora_rank=0` → `FORWARD_ONLY` mismatch → `400 Bad Request`.

> **Base branch is #371.** Once #371 merges to `main`, rebase and re-target this PR to `main`.

## Fix 1 — DataLoader streaming + AppendOnlyPickleLog ref cache

On a 110K-pair, 256K-context dataset the original eager DPO data flow peaked at:

  - `tokenized_pairs` (in-memory list): ~150-250 GiB
  - `ref_cache` (in-memory dict, multi-epoch): ~440 GiB

Both well past the ~120 GiB allocatable on a single CPU node. With this PR peak RAM is bounded by `num_workers * per-worker render footprint` (low tens of GiB) regardless of dataset size or epoch count.

### Phase 1 — DataLoader-driven streaming render

- Reuse `JsonlRenderDataset` + `make_render_dataloader` from #371. The render fn (`_render_pair_worker`) now does schema normalisation, pair rendering, and over-length filtering in one pass — all drop reasons collapse into a `None` return that the loader's collate filters out per batch.
- **No separate render-to-disk phase.** Epoch 0's producer drives the DataLoader directly: tokenize in spawn workers → ref forward (with semaphore-bounded concurrency) → enqueue for the trainer.
- The previous tokenized-pair `DiskBackedDatumStore` is gone entirely.

### Phase 2 — `AppendOnlyPickleLog` for multi-epoch ref cache

For `epochs > 1` the producer additionally appends each enriched pair (`datum + ref_chosen + ref_rejected`) to a tiny disk log so the reference trainer can be released after epoch 0 and epochs 1+ stream the cache from disk.

`AppendOnlyPickleLog` is ~30 LoC of `pickle.dump`/`pickle.load`: sequential append while writing, sequential iteration after `close_write()`, no offset table, no random access. That's all the access pattern needs (epochs 1+ iterate the cache once, front-to-back).

`ref_chosen` / `ref_rejected` are `array.array('f', ...)` (~7x smaller on disk than `list[float]`; `torch.tensor` and `pickle` both handle them natively).

### Architecture

```mermaid
flowchart LR
  subgraph Disk
    JSONL[(dataset.jsonl)]
    RC[(ref_cache.pkl<br/>AppendOnlyPickleLog<br/>multi-epoch only)]
  end

  JSONL --> DS[JsonlRenderDataset<br/>lazy offset table]
  DS --> DL[DataLoader<br/>spawn × N workers<br/>persistent_workers]
  DL -- batch of rendered pairs --> P[ref_producer<br/>asyncio]
  P -- ref_forward<br/>array.array f --> Q{{asyncio.Queue}}
  Q --> C[trainer<br/>forward_backward + optim_step]
  P -. multi-epoch only .-> RC

  RC -- sequential read --> E1[epochs 1+]
  E1 --> C

  classDef store fill:#fef3c7,stroke:#92400e
  class RC store
```

### What changed vs the previous attempt on this branch

The first iteration of this PR built a bespoke `DiskBackedDatumStore` + `multiprocessing.Pool.imap` render pipeline. After landing the same problem in SFT (#371), we converged on a single shared streaming primitive (`JsonlRenderDataset` + DataLoader) so this PR now reuses it. Net delta: `dpo_loop.py` shrinks from 728 → 561 LoC, the tokenized-pair disk store is eliminated, and `MemTracer` rendering hooks are gone (RSS is bounded by construction).

### Root cause

For LoRA DPO (`lora_rank > 0`) the gateway auto-resolves both the policy and reference shapes to `LORA_TRAINER` mode. The cookbook's `_request_trainers` was hardcoding `lora_rank=0` for the reference trainer, which made the gateway infer `FORWARD_ONLY` mode — mismatch with the resolved `LORA_TRAINER` shape → `400 Bad Request` on ref trainer creation.

### Fix

Propagate the actual `lora_rank` to the reference trainer request and the reference `ReconnectableClient`. With `lora_rank > 0`, an explicit `ref_training_shape_id` now intentionally means "provision a separate forward-only LoRA reference trainer on its own GPUs"; the old base-only shared-session reference semantics remain available by leaving `ref_training_shape_id` unset so `setup_infra` uses `policy.create_base_reference()`. Validated end-to-end against `qwen3p5-9b` LoRA DPO on B200 in dev (`fireworks-e2e-tests` account).

## Test plan

- [x] `pytest training/tests/unit/test_streaming.py training/tests/unit/test_dpo_loop.py` — 39 tests, all pass.
  - Adds `AppendOnlyPickleLog` round-trip / ordering / lifecycle coverage.
  - Rewrites `test_dpo_loop.py` against the new render fn (with schema normalisation), the `JsonlRenderDataset` + `AppendOnlyPickleLog` train loop, and the empty-batch skip path.
- [x] Full unit suite: 360 passed, 22 pre-existing failures (SDK `skip_validations` mismatch + unrelated infra/preflight tests). Same 22 failures as the #371 baseline — no new regressions.
- [x] DPO E2E in dev (`qwen3p5-9b` LoRA, 4 × B200, `fireworks-e2e-tests`) completed successfully on the previous iteration of this branch; the DataLoader rewrite is a same-output refactor (producer order is preserved end-to-end so multi-epoch ref-cache replay is bit-for-bit identical).


training/recipes/dpo_loop.py
```diff
@@ -381,23 +405,35 @@ def _run_train_step(epoch: int, step_pairs: list[dict[str, Any]]) -> None:
...
+        # The DataLoader is synchronous; pull each batch via to_thread so
+        # render workers and ref forwards can overlap on the event loop.
+        loader_iter = iter(loader)
+        sentinel = object()
+        while True:
+            batch = await asyncio.to_thread(next, loader_iter, sentinel)
+            if batch is sentinel:
+                break
+            if not batch:  # all rows in this DataLoader batch rendered to None
+                continue
```
renfeichen-fw: **Silent zero-steps regression.** If `_render_pair_worker` filters every pair (too-long, schema drift, bad preferences), every batch is empty, the loop runs to completion, and the job reports COMPLETED with 0 steps and no output model — customer is billed for a no-op.

The pre-PR code raised `"No valid pairs after tokenization"`. Please preserve that contract: count non-None pairs and raise at epoch end (or job end) if the total is 0.

training/recipes/dpo_loop.py
```diff
@@ -547,33 +590,43 @@ def _on_trainer_status(msg: str) -> None:
...
+
+        max_seq_len = infra.max_seq_len
+        init_args = (cfg.tokenizer_model, cfg.renderer_name, max_seq_len)
+        # Initialise in the parent so the num_workers <= 1 fallback uses
+        # the same render_fn as the spawn workers.
+        _init_pair_worker(*init_args)
+        worker_init_fn = functools.partial(_init_pair_worker, *init_args)
+
+        pair_dataset = JsonlRenderDataset(
+            cfg.dataset, _render_pair_worker, max_examples=cfg.max_pairs,
```
renfeichen-fw: **`max_pairs` semantic change.** Before: cap on *valid rendered pairs*. Now: cap on *raw JSONL rows*, before filtering.

Customers who set `max_pairs=N` expecting exactly `N` pairs will get fewer actual training pairs once any row is filtered out. Either rename to `max_rows` or restore post-filter semantics by capping inside the consumer loop.

training/tests/unit/test_dpo_loop.py
```diff
@@ -229,225 +210,344 @@ def forward_backward_custom(self, datums, loss_fn):
...
+        cfg = module.Config(
+            log_path=str(tmp_path), epochs=1, batch_size=2, render_workers=0,
+        )
+        step = asyncio.run(
+            module._train_loop(
+                ds, None, CountingReference(), _FakePolicy(),
+                adam_params={"lr": 1e-4}, cfg=cfg, step_offset=0,
+            )
+        )
+        assert step == 0
```
renfeichen-fw: This test locks in the silent zero-steps behavior (`assert step == 0`). If/when we fix that (hard-fail on all-filtered dataset), this test will break and the fix will get reverted.

Suggest flipping to `pytest.raises(RuntimeError, match="No valid pairs")` so the test encodes the customer-safe contract instead.

renfeichen-fw (changes requested):
Same silent-zero-steps pattern as #371, plus a subtle `max_pairs` semantic change and a test that codifies the regression. See inline.

training/recipes/dpo_loop.py
```diff
@@ -547,33 +590,43 @@ def _on_trainer_status(msg: str) -> None:
...
+        )
+        if len(pair_dataset) == 0:
+            raise RuntimeError(f"No data found in {cfg.dataset}")
 
-        tokenizer = transformers.AutoTokenizer.from_pretrained(cfg.tokenizer_model, trust_remote_code=True)
-        renderer = build_renderer(tokenizer, cfg.tokenizer_model, cfg.renderer_name)
         logger.info(
-            "Using renderer=%s for preference tokenization",
+            "Streaming %d preference rows from %s (renderer=%s, workers=%d)",
```
renfeichen-fw: **Rendering progress dropped.** Pre-PR `_tokenize_pairs` called `runner.report_rendering_progress` at ~5 % intervals and logged the filtered-pair count. Both are gone — same gap as the SFT PR (#371).

DPO epoch 0's DataLoader → ref-forward pipeline can run longer than SFT's render-only phase (each batch blocks on a ref forward RPC). Consider emitting progress inside `_ref_producer` — it already knows `len(pair_dataset)` and can log at coarse intervals (e.g. every ~5 % of batches consumed) without adding latency to the critical path.

training/utils/infra.py
```diff
@@ -1182,11 +1182,11 @@ def _resolve_reference_shape(
     """Return (ref_profile, resolved_ref_shape_id)."""
     if ref_training_shape_id:
         if lora_rank > 0:
-            logger.warning(
-                "ref_training_shape_id=%r is set alongside lora_rank=%d. This provisions a "
-                "separate full-param forward-only reference trainer, which defeats LoRA's "
-                "GPU savings. For LoRA, leave ref_training_shape_id unset — setup_infra "
-                "will use the shared-session reference (policy.create_base_reference()).",
+            logger.info(
+                "Using explicit reference shape %s alongside lora_rank=%d. The ref "
```
renfeichen-fw, line 1207: **Silent semantic flip for existing `ref_training_shape_id` + LoRA configs.**

On `main`, `_request_trainers` / `_make_reference_client` hard-code `lora_rank=0`, so any LoRA policy with an explicit ref shape got a full-param forward-only reference. This PR now passes the policy's `lora_rank` through without validating that the supplied shape is LoRA-capable.

Two problems:
1. **Backward compat.** Existing configs pointing `ref_training_shape_id` at a full-param shape while running LoRA flip from "works, extra GPUs" to a trainer-mode / shape mismatch at create time.
2. **DPO correctness.** The DPO reference is normally the *pre-adapter* distribution (base or SFT-warmstarted). Loading the policy's LoRA adapter into the ref changes the reference distribution — the KL target is now base+adapter, not base. If that's intentional, please say so in the PR description + a code comment; otherwise this should stay at `lora_rank=0` for the forward-only ref.

At minimum: validate the supplied ref shape is LoRA-capable when propagating `lora_rank > 0`, and fail loudly rather than at the gateway.

renfeichen-fw (changes requested):
One more concern on the LoRA ref-trainer propagation: backward-incompat + potential DPO-semantics change. See inline.

training/recipes/dpo_loop.py
xiaoyifan: Fixed in `ece01d6`: `_train_loop` now tracks rendered/usable pairs, logs the filtered count, and raises `RuntimeError("No valid pairs after tokenization")` when epoch 0 produces no valid data instead of completing at step 0.

training/recipes/dpo_loop.py
xiaoyifan: Fixed in `ece01d6`: `max_pairs` is back to counting valid rendered pairs, not raw JSONL rows. The producer now caps after filtering and also re-batches valid pairs so filtered rows don't silently change optimizer-step sizing.

training/tests/unit/test_dpo_loop.py
xiaoyifan: Updated in `ece01d6`: the regression test now asserts `pytest.raises(RuntimeError, match="No valid pairs after tokenization")` so we keep the customer-safe contract instead of locking in the old step-0 behavior.

training/recipes/dpo_loop.py
xiaoyifan: Fixed in `ece01d6`: `_ref_producer` now emits coarse `runner.report_rendering_progress(...)` updates during epoch 0 and restores the filtered-row logging once the producer finishes.

training/utils/infra.py
xiaoyifan: Kept the LoRA propagation intentionally, but clarified the contract. For this branch / CP behavior we do want explicit `ref_training_shape_id` + `lora_rank > 0` to mean a separate forward-only LoRA reference trainer (own GPUs), while the old base-only shared-session reference path is still available by leaving `ref_training_shape_id` unset. I also updated the PR description to say that explicitly, and the inline comments in `infra.py` still call out the shared-session alternative.

PR #376, opened 2026-04-23
feat(rl): multi-minibatch PPO loop with single prox snapshot

Hecate0821:
## Summary
- Hoist `prox_lp = policy.forward(data, "cross_entropy")` above the training loop so one snapshot serves all inner optim steps.
- Add `Config.ppo_n_minibatches` (default **1**, preserves legacy 1:1 semantics byte-for-byte). At K>1 the policy drifts across inner steps and the PPO ratio measures real drift — matches AReaL's `ppo_update` and Slime's per-rollout `compute_log_prob` + minibatch loop.
- Move the DCP save to a rollout boundary check so `data_consumed` stays K-independent on resume.
- Rescale `total_rl_steps` by `ppo_n_minibatches` so the runner progress bar stays in optim-step units.

## Context
Today cookbook computes `prox_lp` inside `fwd_bwd_one` and runs a single fwd_bwd + optim_step per rollout → `prox_lp ≡ resp_pi` byte-for-byte → ratio ≡ 1 → PPO clip never fires. This is half of decoupled-PPO wired up with no consumer. AReaL/Slime run decoupled-PPO with multi-minibatch inner loops (or off-policyness) so the ratio is meaningful. This PR makes cookbook match.

## Algorithm diff

| | Before (K=1) | After (K>1) |
|---|---|---|
| `prox_fwd` per rollout | 1 | 1 |
| `forward_backward` per rollout | 1 | K |
| `optim_step` per rollout | 1 | K |
| PPO ratio at inner step k | `exp(π−π_prox)≈1` | drifts with k |
| PPO clip | no-op | activates |

## Behavioral guarantees
- `ppo_n_minibatches=1` → identical control flow to prior code: single `fwd_bwd_minibatch` result, single `optim_step`, metrics byte-identical (the averaging fix reduces to mean-of-one).
- DCP save now only fires at rollout boundaries (checking `rollouts_completed % dcp_save_interval`), not mid-rollout. Fixes an interval-divisibility bug where e.g. K=4, interval=10 would never trigger.

## Correctness fixes folded in
Reviewer flagged three latent K>1 mis-accountings introduced by the refactor. All three sit in code paths the single-shape CI never walks (DPO owns the only DCP-resume path; `save_final_checkpoint=False` in the test) so unit tests are the guardrail. All fixed in one commit, inlined at the callsites.

1. **Resume skipped K× rollout batches.** `step_offset` is optim-step units; `rl_dataset` is indexed per rollout batch. Divide by `max(1, ppo_n_minibatches)` before indexing.
2. **Final-checkpoint `data_consumed` over-counted by K×.** In-loop save used `rollouts_completed`; the mirror final save below `main()` still used `(global_step - step_offset) * pgps`. Now matches the in-loop formula (divide optim-step delta by K before multiplying by pgps).
3. **Per-step `train/*` metrics only reflected the last minibatch.** `compute_step_metrics` took `fwd_bwd_results[-1]`; now averages across all minibatches. K=1 reduces to the single-result case (mean == that result), byte-for-byte equivalent to the pre-PR code.

## Cosmetic
Rename: `K → num_minibatches`, `k_idx → minibatch_idx` inside `train_step` — matches surrounding vocabulary (`minibatch_size`, `minibatch_start`, `minibatch_end`).

## Test plan
- [x] Unit: K=1 byte-equivalence — `TestFwdBwdMinibatchAveraging::test_k1_matches_single_fwd_bwd_behavior` covers the averaging path; the resume-rollout and data_consumed fixes are trivially K=1-equivalent (divide-by-1 noop).
- [x] K=2 behavioral — single-shape CI against `qwen3-4b-minimum-lora` with `policy_loss="reinforce"` (drift-clip coverage via unit test, not CI): PASS.
- [ ] Regression run with stress test enabled: in progress (v8 / fireworks run 24819413563).

### DIFF
--- a/training/recipes/rl_loop.py
+++ b/training/recipes/rl_loop.py
@@ -24,11 +24,12 @@ from __future__ import annotations
 import os
 import re
 import json
+import math
 import signal
 import asyncio
 import logging
 from contextlib import ExitStack
-from typing import List, Optional
+from typing import Any, List, Optional
 from dataclasses import field, dataclass
 from concurrent.futures import ThreadPoolExecutor
 
@@ -145,6 +146,14 @@ class Config:
     """Asymmetric upper clip bound (GRPO only)."""
     ratio_log_cap: float = 20.0
     """Log-ratio clamp for ``policy_loss="importance_sampling"``."""
+    ppo_n_minibatches: int = 1
+    """Number of inner PPO minibatches per rollout batch.
+
+    Each rollout batch runs one ``prox_fwd`` snapshot followed by
+    ``ppo_n_minibatches`` × (``forward_backward`` + ``optim_step``). When
+    >1, the policy drifts across inner steps, so ``prox_lp`` anchors the
+    PPO ratio and the clip does real work. ``1`` reproduces the legacy
+    1:1 behavior."""
     tis: TISConfig = field(default_factory=TISConfig)
     """TIS (Train-Inference IS) weight correction config."""
 
@@ -622,23 +631,15 @@ def main(
             eps_clip_high=cfg.eps_clip_high,
         )
 
-        def fwd_bwd_one(prompt_groups: list[PromptGroup]):
-            """One minibatch update using the builtin or client-side loss path."""
-            if not prompt_groups:
-                raise ValueError("fwd_bwd_one requires at least one prompt group")
+        def fwd_bwd_minibatch(
+            data, adv, ref_lp, prompt_lens, inf_lp, prox_lp,
+        ):
+            """One inner PPO minibatch using builtin or client-side loss path.
 
-            data, adv, ref_lp, prompt_lens, inf_lp = combine_prompt_groups(prompt_groups)
-
-            t0 = _time.time()
-            prox_fwd = policy.forward(data, "cross_entropy")
-            prox_lp = [prox_fwd.loss_fn_outputs[i]["logprobs"].data for i in range(len(data))]
-            logger.info("policy_forward: done (%.1fs)", _time.time() - t0)
-
-            t0 = _time.time()
+            Callers pre-compute ``prox_lp`` once per rollout batch and pass
+            a slice of the flattened rollout tensors for this minibatch.
+            """
             if builtin_server_loss is not None:
-                # Server-side builtin path: pre-pack the rollout tensors into
-                # datums the trainer kernel understands, then call
-                # forward_backward(...).
                 kernel_loss, kernel_config = builtin_server_loss
                 rl_datums = build_builtin_loss_datums(
                     data,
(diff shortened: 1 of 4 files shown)


training/recipes/rl_loop.py
```diff
@@ -769,7 +807,7 @@ def _loop_metrics_callback(loop_metrics: dict) -> None:
         for i_step in range(step_offset, len(rl_dataset)):
             remaining_rows.extend(rl_dataset.get_batch(i_step))
 
-        total_rl_steps = len(rl_dataset) - step_offset
+        total_rl_steps = len(rl_dataset) * max(1, cfg.ppo_n_minibatches) - step_offset
```
renfeichen-fw: **Bug: resume with `ppo_n_minibatches > 1` skips K× too many rollout batches.**

`step_offset = resume_info.step` is saved in **optim-step units** (lines 719 & 738 now increment `step` by K per rollout), but `rl_dataset` is indexed in **rollout-batch units** — `len(rl_dataset) = ceil(len(rows) / prompt_groups_per_step)` (see `training/utils/data.py:34`).

Concrete failure case: K=2, save at rollout 3 → stored `step = step_offset_initial + 6`. Resume from that checkpoint → `step_offset = 6` → this loop iterates `range(6, len(rl_dataset))` and throws away 6 rollout batches of training data instead of 3.

K=1 is unaffected because optim-steps == rollout-batches.

### Suggested fix

```python
K = max(1, cfg.ppo_n_minibatches)
resume_rollouts = step_offset // K
remaining_rows = []
for i_step in range(resume_rollouts, len(rl_dataset)):
    remaining_rows.extend(rl_dataset.get_batch(i_step))
```

(`step_offset` is guaranteed to be a multiple of K because the in-loop DCP save at line 727 only fires at rollout boundaries — so `step_offset // K` is exact, not lossy.)

### Unit test

Extract a tiny pure helper and unit-test it:

```python
def _resume_rollout_index(step_offset: int, ppo_n_minibatches: int) -> int:
    """Convert a saved optim-step counter back to a rollout-batch index."""
    return step_offset // max(1, ppo_n_minibatches)
```

Then in `training/tests/unit/test_rl_loop.py`:

```python
import pytest
import training.recipes.rl_loop as module

@pytest.mark.parametrize("step_offset, K, expected", [
    (0,  1, 0),   # fresh
    (5,  1, 5),   # K=1 resume (unchanged from today)
    (0,  2, 0),
    (6,  2, 3),   # 3 rollouts × 2 optim steps
    (12, 4, 3),   # 3 rollouts × 4 optim steps
])
def test_resume_rollout_index_converts_optim_steps_to_rollouts(step_offset, K, expected):
    assert module._resume_rollout_index(step_offset, K) == expected
```

Plus a one-shot integration assertion with a stub dataset:

```python
def test_main_resume_starts_from_correct_rollout_with_k_gt_1(monkeypatch, tmp_path):
    # Build an RLPromptDataset of 10 rollout batches, monkeypatch resolve_resume
    # to return step=6, data_consumed=..., and run up to the sample_one_prompt
    # iterator construction. Assert len(remaining_rows) == 4 * prompt_groups_per_step,
    # not 4.
```

training/recipes/rl_loop.py
```diff
@@ -649,69 +650,106 @@ def fwd_bwd_one(prompt_groups: list[PromptGroup]):
...
+            if dcp_interval > 0 and rollouts_completed > 0 and rollouts_completed % dcp_interval == 0:
                 logger.info("[step %d] dcp_save...", step)
                 t0 = _time.time()
-                logger.info("[step %d] dcp_save: done (%.1fs)", step, _time.time() - t0)
-                _data_consumed = (resume_info.data_consumed if resume_info else 0) + (
-                    step - step_offset
-                ) * prompt_groups_per_step
+                data_consumed = (resume_info.data_consumed if resume_info else 0) + (
+                    rollouts_completed * prompt_groups_per_step
+                )
```
renfeichen-fw, line 732: **Bug: the mirror final-checkpoint save at the bottom of `main(...)` was not updated for `K > 1`.**

The in-loop save here is correct — it divides by K via `rollouts_completed`. But the final checkpoint a few dozen lines below still uses the pre-PR formula:

```python
# training/recipes/rl_loop.py (unchanged in this PR) — lines 829-833
if cfg.save_final_checkpoint and global_step > step_offset:
    try:
        _data_consumed = (resume_info.data_consumed if resume_info else 0) + (
            global_step - step_offset
        ) * prompt_groups_per_step
```

`global_step - step_offset` is the number of **optim steps** executed in this run, not rollout batches. At K=2 it over-counts `data_consumed` by 2×, and that wrong value is what gets serialized into the final checkpoint metadata — so the next resume reads a corrupt `data_consumed`, and if bug #1 is also unfixed, you dataset-skip *that* K× as well. Two bugs compound.

### Suggested fix

Introduce one helper and use it in **both** places (in-loop DCP save _and_ final save):

```python
def _compute_data_consumed(
    *,
    resume_data_consumed: int,
    global_step: int,
    step_offset: int,
    prompt_groups_per_step: int,
    ppo_n_minibatches: int,
) -> int:
    K = max(1, ppo_n_minibatches)
    rollouts_in_this_run = (global_step - step_offset) // K
    return resume_data_consumed + rollouts_in_this_run * prompt_groups_per_step
```

Then the in-loop save becomes:

```python
data_consumed = _compute_data_consumed(
    resume_data_consumed=resume_info.data_consumed if resume_info else 0,
    global_step=step,
    step_offset=step_offset,
    prompt_groups_per_step=prompt_groups_per_step,
    ppo_n_minibatches=cfg.ppo_n_minibatches,
)
```

…and the final save at line 831 uses the same helper with `global_step=global_step`. That's the core win of extracting — one formula, two callsites, impossible to drift again.

### Unit test

```python
import pytest
import training.recipes.rl_loop as module

@pytest.mark.parametrize("resume, step, offset, pgps, K, expected", [
    # K=1: rollouts == optim steps. Must match pre-PR behavior byte-for-byte.
    (0,   5, 0,  1, 1, 5),
    (100, 7, 2,  4, 1, 100 + 5 * 4),
    # K=2: rollouts == optim_steps / 2.
    (0,   6, 0,  1, 2, 3),
    (0,   6, 0,  4, 2, 12),
    # K=4, resuming from a prior checkpoint:
    (100, 20, 12, 1, 4, 100 + 2),  # 2 rollouts in this run on top of 100 prior
    (100, 20, 12, 4, 4, 100 + 8),
])
def test_compute_data_consumed(resume, step, offset, pgps, K, expected):
    assert module._compute_data_consumed(
        resume_data_consumed=resume,
        global_step=step,
        step_offset=offset,
        prompt_groups_per_step=pgps,
        ppo_n_minibatches=K,
    ) == expected


def test_compute_data_consumed_k1_matches_pre_pr_formula():
    """K=1 must be byte-identical to the legacy formula."""
    resume, step, offset, pgps = 100, 7, 2, 4
    legacy = resume + (step - offset) * pgps
    new = module._compute_data_consumed(
        resume_data_consumed=resume,
        global_step=step,
        step_offset=offset,
        prompt_groups_per_step=pgps,
        ppo_n_minibatches=1,
    )
    assert new == legacy
```

The "K=1 matches legacy" test is the cheap way to make sure refactoring doesn't regress the existing shipping path.

training/recipes/rl_loop.py
```diff
@@ -649,69 +650,106 @@ def fwd_bwd_one(prompt_groups: list[PromptGroup]):
...
             metrics = compute_step_metrics(
                 prompt_groups=prompt_groups,
-                fwd_bwd_results=[fwd_bwd_result],
+                fwd_bwd_results=fwd_bwd_results,
                 optim_result=optim_result,
-                n_accum=1,
+                n_accum=len(fwd_bwd_results),
                 timing_metrics=flush_timing(),
                 loop_stats=loop_stats,
                 completions_per_prompt=completions_per_prompt,
```
renfeichen-fw, line 755: **Nit / soft issue: per-step `train/*` metrics only reflect the _last_ minibatch under K > 1.**

`compute_step_metrics` picks `fwd_bwd_results[-1]` for the `train/*` keys and takes the last `optim_result` only:

```python
# training/utils/rl/metrics.py lines 117-126
if optim_result and hasattr(optim_result, "metrics") and optim_result.metrics:
    for k, v in optim_result.metrics.items():
        if k not in _SKIP_REMOTE_KEYS:
            metrics[f"train/{k}"] = v

last_fwd_bwd = fwd_bwd_results[-1] if fwd_bwd_results else None
if last_fwd_bwd is not None:
    for k, v in last_fwd_bwd.metrics.items():
        if k not in _SKIP_REMOTE_KEYS:
            metrics[f"train/{k}"] = v
```

For `K > 1`, the headline claim of this PR is that the PPO clip now activates. But `ppo_clip_frac`, `ppo_ratio_mean`, and grad norms in wandb will only ever show the final minibatch — i.e. the minibatch where the ratio has drifted the most from 1. Early minibatches (where ratio is still ≈ 1) won't show up at all. This:
- Under-reports total clip activity across the rollout batch.
- Makes the "byte-for-byte K=1 parity" claim hard to verify empirically (since `K=1` hits `fwd_bwd_results[-1]` which is the _only_ element, matching old behavior — good), but for `K>1` you're reporting only the most aggressive minibatch.

### Suggested fix

Average the per-minibatch metrics across `fwd_bwd_results`, with K=1 reducing to the current behavior. Small change in `compute_step_metrics` gated on `len(fwd_bwd_results) > 1`, or just always-average (averaging one value is a no-op).

### Unit test

Extend `training/tests/unit/test_rl_metrics_aliases.py`:

```python
from unittest.mock import MagicMock
from training.utils.rl.metrics import compute_step_metrics


def _fake_fwd_bwd(clip_frac, ratio_mean):
    m = MagicMock()
    m.metrics = {"ppo_clip_frac": clip_frac, "ppo_ratio_mean": ratio_mean}
    return m


def test_compute_step_metrics_averages_ppo_clip_across_minibatches():
    fwd_bwds = [
        _fake_fwd_bwd(clip_frac=0.0, ratio_mean=1.00),  # mb 1: ratio ~ 1
        _fake_fwd_bwd(clip_frac=0.1, ratio_mean=1.05),
        _fake_fwd_bwd(clip_frac=0.3, ratio_mean=1.20),  # mb K: most drift
    ]
    metrics = compute_step_metrics(
        prompt_groups=[],
        fwd_bwd_results=fwd_bwds,
        optim_result=None,
        n_accum=len(fwd_bwds),
        timing_metrics={},
    )
    assert metrics["train/ppo_clip_frac"] == pytest.approx((0.0 + 0.1 + 0.3) / 3)
    assert metrics["train/ppo_ratio_mean"] == pytest.approx((1.0 + 1.05 + 1.2) / 3)


def test_compute_step_metrics_k1_matches_single_fwd_bwd_behavior():
    """K=1 must report the single minibatch's metrics directly — no drift from prior behavior."""
    only = _fake_fwd_bwd(clip_frac=0.42, ratio_mean=1.07)
    metrics = compute_step_metrics(
        prompt_groups=[], fwd_bwd_results=[only], optim_result=None,
        n_accum=1, timing_metrics={},
    )
    assert metrics["train/ppo_clip_frac"] == 0.42
    assert metrics["train/ppo_ratio_mean"] == 1.07
```

Not blocking — the algorithm is correct either way. But without this change the PR author's own test plan ("confirm `ppo_clip_frac > 0` on minibatch 2") is basically impossible to verify through wandb, because minibatch 2's `ppo_clip_frac` is what _replaces_ minibatch 1's in the current code, rather than being reported alongside.

renfeichen-fw (commented):
Thanks for the fix — hoisting `prox_fwd` out of the inner loop is the right call, and the `ppo_n_minibatches=1` backward-compat path looks clean.

Two real bugs remain in the **step accounting**. They don't affect `K=1` runs, but they do break the `K>1` story, and one of them compounds across resumes:

1. **Resume bug** — `for i_step in range(step_offset, len(rl_dataset))` indexes `rl_dataset` in rollout-batch units but `step_offset` is now in optim-step units (= K × rollouts). At `K=2`, resume from step 6 skips 6 rollout batches instead of 3.
2. **Final-checkpoint `_data_consumed`** — the in-loop DCP save was updated to divide by K (via `rollouts_completed`), but the mirror at the end of `main(...)` still multiplies `(global_step - step_offset) * prompt_groups_per_step`, over-counting by a factor of K. This poisons `data_consumed` in the last-step checkpoint and then compounds bug 1 on the next resume.

Inline comments below with the fixes and suggested unit tests. I'd also recommend extracting two small pure helpers so these become trivially testable without spinning up a real trainer — see the "Test plan" section at the bottom.

---

### Test plan (recommendation)

The two bugs above both live inside closures inside `main(...)`, so they're not directly unit-testable today. The cheapest path is to extract two pure helpers and cover them in `training/tests/unit/test_rl_loop.py` (same test file the PR author already has for this recipe). Concrete suggestions in the inline comments.

Once the helpers are in place, the test coverage I'd want to see is:

- `K=1, step_offset=0` → behavior identical to `main`.
- `K=1, step_offset=N` → resume skips exactly N rollout batches; final `data_consumed` advances by N × prompt_groups_per_step.
- `K=2, step_offset=0, 3 rollouts run` → final `data_consumed = 3 * prompt_groups_per_step`, not `6 * prompt_groups_per_step`.
- `K=2, step_offset=6 (resumed from rollout 3)` → dataset iteration starts at rollout index 3, not 6.
- `K=4, step_offset=12, 2 more rollouts run` → final `data_consumed` advances by 2 × prompt_groups_per_step on top of the resumed `data_consumed`.

The metrics concern (comment 3) is softer — an assertion in `test_rl_metrics_aliases.py` or similar that `ppo_clip_frac` is averaged across minibatches would be nice but isn't a correctness blocker.

renfeichen-fw (approved):

PR #381, opened 2026-04-23
refactor(rl): extract infra helpers from rl_loop.py

Hecate0821:
## Summary

Move pure-infra plumbing out of \`recipes/rl_loop.py\` into \`training/utils\` so the recipe reads as training logic rather than setup boilerplate. **rl_loop.py: 907 → 809 lines (-98).**

What was moved:

| From rl_loop.py | To |
|---|---|
| \`_dump_trajectory\` | \`training/utils/rl/trajectory.py::dump_trajectory_jsonl\` |
| SIGTERM/SIGINT handler install | \`training/utils/runner.py::install_cleanup_signal_handlers\` |
| DCP save-if-due block inside \`train_step\` | \`training/utils/checkpoint_utils.py::save_dcp_checkpoint_if_due\` |
| Final-checkpoint + promote block | \`training/utils/checkpoint_utils.py::save_final_checkpoint_and_promote\` |
| Adaptive concurrency builder + echo-log | \`training/utils/config.py::build_adaptive_concurrency_controller\` |

Also:
- Collapsed the duplicate timing logs in \`_weight_sync\` (already wrapped in \`timer(\"weight_sync\")\`).
- Collapsed \`TrainerJobManager\`/\`DeploymentManager\` default construction via shared kwargs dict.
- Dropped now-unused \`signal\` / \`json\` imports.

No behavior change: each extracted helper preserves the original log format, error message, checkpoint kind/path, and status writes.

## Test plan

- [ ] \`pytest training/tests/unit\` — should all pass (no public API touched beyond the new helpers).
- [ ] Run an RL smoke (e.g. \`python -m recipes.rl_loop\` with a short dataset) and confirm: trajectory JSONL still written, DCP checkpoint still fires on \`dcp_save_interval\`, final checkpoint still saved and promoted when \`output_model_id\` is set, signal handling still triggers cleanup on SIGTERM.

### DIFF
--- a/training/recipes/rl_loop.py
+++ b/training/recipes/rl_loop.py
@@ -23,9 +23,7 @@ from __future__ import annotations
 
 import os
 import re
-import json
 import math
-import signal
 import asyncio
 import logging
 from contextlib import ExitStack
@@ -38,7 +36,7 @@ import transformers
 
 from fireworks.training.sdk import DeploymentManager, TrainerJobManager
 from fireworks.training.sdk.client import GradAccNormalization
-from fireworks.training.sdk.deployment import AdaptiveConcurrencyController, DeploymentSampler
+from fireworks.training.sdk.deployment import DeploymentSampler
 from fireworks.training.sdk.weight_syncer import WeightSyncer
 from training.utils import (
     DEFAULT_ADAM,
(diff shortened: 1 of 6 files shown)


Hecate0821:
Closing — the extractions don't help readability enough to justify the indirection.

PR #386, opened 2026-04-24
fix(supervised): preserve thinking on every assistant turn for Kimi/Qwen3/DeepSeek-thinking

benjibc:
## Problem

After PR #378 fixed the `reasoning_content` alias, customers reported fine-tuned Kimi K2.5 models still emit an empty `<think></think>` on ~50% of responses at inference. Adding `reasoning_content` into the literal `content` as a `<think>...`` workaround restores CoT on all responses.

## Root cause

KimiK2/K2.5/K2.6, Qwen3-thinking, and DeepSeek V3-thinking renderers **strip `<think>` from every historical assistant turn** and only keep it on the last assistant turn (`has_extension_property=False`).

The SFT loop rendering path called `renderer.build_supervised_example` (singular) with `train_on_what=all_assistant_messages`. For an N-turn conversation, the trained sequence looked like:

```
turn 0..N-2:  <think></think>{answer}<|im_end|>
turn N-1:     <think>{cot}</think>{answer}<|im_end|>
```

even when **every turn** in the dataset had a non-empty `reasoning_content`. The fine-tuned model then directly learned to emit `<think></think>` for the majority of turns.

### Direct reproduction (Kimi K2.5 tokenizer + actual renderer)

Two-assistant conversation; both turns have `reasoning_content`:

```
BEFORE: render_messages_to_datum (singular)
  trained run 0: '<think></think>4<|im_end|>'                         <-- FIRST CoT DROPPED
  trained run 1: '<think>SECOND THINKING</think>6<|im_end|>'          <-- only last CoT kept

AFTER: render_messages_to_datums (plural)
  example 0 run 0: '<think>FIRST THINKING</think>4<|im_end|>'         <-- kept
  example 1 run 0: '<think>SECOND THINKING</think>6<|im_end|>'        <-- kept
```

This matches the customer's '`reasoning_content` added to content' workaround: literal-text thinking survives because the stripping only touches `ThinkingPart`.

## Fix

Add `render_messages_to_datums` (plural) which calls `renderer.build_supervised_examples` (plural). The plural method is already implemented on each thinking renderer:

- **KimiK2Renderer** splits multi-turn conversations into one example per user turn so every CoT is preserved in the turn it is trained on (kimi_k2.py#335).
- **Base renderer** (`has_extension_property=True` — Llama3, RoleColon, Qwen3 non-thinking, DeepSeek non-thinking) returns a singleton list, so behavior for plain instruct models is **unchanged**.

`sft_loop.py` `_map_fn` now returns a list, and the training/eval loops `extend` into their datum lists.

`render_messages_to_datum` (singular) is retained as a back-compat shim for paths that explicitly need a single datum (preference pairs, token-mask eval, etc.).

## Tests

- `test_render_messages_to_datums_uses_plural_build_supervised_examples`: plural form must delegate to `build_supervised_examples`.
- `test_render_messages_to_datums_filters_normalized_reasoning_content`: `reasoning_content` still flows through `normalize_messages` before rendering.
- `test_render_messages_to_datum_still_uses_singular_for_back_compat`: singular shim is unchanged.
- Existing `test_sft_loop` tests were updated to patch the plural name and wrap outputs in a list; `StubRenderer` now implements both singular and plural forms so extension-property renderers keep working.

Full unit test suite: `374 passed, 53 skipped` in `training/tests/unit/`.

## Surfaces

- `training/utils/supervised.py`: new public `render_messages_to_datums`; shared `_datum_from_rendered_example` helper.
- `training/utils/__init__.py`: re-export `render_messages_to_datums`.
- `training/recipes/sft_loop.py`: switch SFT training+eval rendering to the plural API.
- `training/tests/unit/test_supervised_rendering.py`: 3 new tests.
- `training/tests/unit/test_sft_loop.py`: update monkeypatches to use the plural name; extend `StubRenderer`.

### DIFF
--- a/training/recipes/sft_loop.py
+++ b/training/recipes/sft_loop.py
@@ -50,7 +50,7 @@ from training.utils import (
     build_renderer,
     parse_train_on_what,
     auto_select_training_shape,
-    render_messages_to_datum,
+    render_messages_to_datums,
     resolve_renderer_name,
 )
 from training.utils.client import DEFAULT_TIMEOUT_S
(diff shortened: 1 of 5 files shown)


benjibc:
Closing — the existing `KimiK2.build_supervised_example` already implements the intended interleave-thinking semantics (preserve `<think>` for assistants after the last non-tool-call assistant, i.e. the last tool-chain). Verified on a 1896-row customer dataset that 100% of `reasoning_content` messages fall inside that preserved suffix, so the singular path already renders them correctly. The plural split introduced here over-corrects — it would force every assistant turn to train with its own CoT preserved, beyond the interleave boundary the renderer intentionally draws. Re-investigating the real cause.

PR #394, opened 2026-04-27
feat(glm5): render assistant tool_calls byte-matching the upstream chat template

zkx06111:
The GLM-5.1 renderer previously dropped the structured `tool_calls` field on assistant messages, so any agentic SFT data using OpenAI-style tool calls silently produced training tokens that omitted the call params -- the model never learned to emit them. This change wires `tool_calls` into `_render_assistant` so calls become part of the assistant's output span, which means `train_on_what=all_assistant_messages` now applies loss to tool-call params alongside regular assistant text. `role: "tool"` responses remain context-only as before.

Format matches `zai-org/GLM-5.1`'s shipped Jinja template byte-for-byte: `<tool_call>{name}<arg_key>k</arg_key><arg_value>v</arg_value>...</tool_call>`, multiple calls concatenated with no separator. String args pass through; non-string args go through `json.dumps(..., ensure_ascii=False)` to mirror the template's `tojson` branch.

Tests cover five tool-call shapes (single call, call-with-text, multi-call, mixed arg types, full agent loop) for byte-parity against `apply_chat_template`, plus a weight-mask check confirming tool-call params land inside trained tokens under `all_assistant_messages`.

### DIFF
--- a/training/renderer/glm5.py
+++ b/training/renderer/glm5.py
@@ -52,12 +52,23 @@ Other invariants:
   ``<|assistant|><think>`` for thinking mode (default),
   ``<|assistant|></think>`` for non-thinking mode.
 
-Only text-only chat, with and without thinking, is implemented here.
-Tool calls and multimodal content are left for a future extension.
+Tool-call layout (assistant turns only — ``role: "tool"`` responses are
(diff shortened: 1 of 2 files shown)


benjibc (approved):
LGTM!

xiaoyifan (approved):

PR #396, opened 2026-04-28
Rename PPO snapshot logprobs

Hecate0821:
## Summary
- Rename the PPO snapshot logprob variable from prox_lp to old_policy_logprobs
- Remove the prox prefix from related docs and logging for the snapshot forward pass
- Use Slime's old-policy meaning while keeping cookbook's logprobs spelling convention

## Verification
- python -m py_compile training/recipes/rl_loop.py
- git diff --check
- targeted diff scan for common secret patterns


websterbei (commented):
much cleaner!

websterbei (approved):

PR #397, opened 2026-04-28
[codex] make GLM5 preserve-thinking SFT default

benjibc:
## Summary
- make the registered `glm5` renderer preserve historical thinking by default, so `ALL_ASSISTANT_MESSAGES` can stay as one supervised datum without adding a second public renderer name
- keep `GLM5Renderer(strip_thinking_from_history=True)` as an explicit strip-history mode and split those non-extension-safe examples by user turn
- route SFT rendering through `render_messages_to_datums` so renderer-produced split examples are used when a non-extension renderer needs them
- add GLM tests mirroring Kimi preserve-thinking, tool-call, and split-example coverage

## Why
Tinker splits Kimi examples when the active renderer strips historical thinking, because earlier assistant turns would otherwise be trained under a different token prefix than generation used at that turn. For GLM interleaved/preserved-thinking data, the better cookbook default is the preserve-thinking contract described by Z.AI: preserve and return historical `reasoning_content` with tool results. That lets `train_on_what=ALL_ASSISTANT_MESSAGES` remain the normal path while still leaving a safe split fallback for explicit strip-history usage.

## Validation
- `training/.venv/bin/python -m pytest training/tests/unit/test_glm5_renderer.py training/tests/unit/test_supervised_rendering.py training/tests/unit/test_sft_loop.py -q`
- `training/.venv/bin/python -m pytest training/tests/unit -q`
- `training/.venv/bin/python -m py_compile training/renderer/glm5.py training/tests/unit/test_glm5_renderer.py`
- `git diff --check`


xiaoyifan (approved):

training/utils/supervised.py
```diff
@@ -748,6 +763,58 @@ def render_messages_to_datum(
...
+        try:
+            examples = list(
+                renderer.build_supervised_examples(
+                    normalized_messages,
+                    train_on_what=effective_train_on_what,
+                )
+            )
+        except (AttributeError, NotImplementedError):
+            examples = [
+                renderer.build_supervised_example(
```
Hecate0821: If base class build_supervised_examples is not implemented, this except would make it silent and render with wrong datum. Need to make it fail fast

training/utils/supervised.py
```diff
@@ -748,6 +763,58 @@ def render_messages_to_datum(
...
+                    normalized_messages,
+                    train_on_what=effective_train_on_what,
+                )
+            ]
+    else:
```
Hecate0821: A multi‑turn row with weight flags rendered through a non‑extension renderer. It should got with the first branch

Hecate0821 (commented):
We need a more clear dispatcher for `render_messages_to_datums`. The if-else are kind of orthogonal.

PR #403, opened 2026-04-29
fix(glm5): mask leading <think> token to match inference generation suffix

morgendave:
## Summary

- Moves the leading `<think>` token from the trained output span (weight=1.0) to the masked header (weight=0.0) in the GLM-5.1 renderer, aligning with the inference generation suffix `<|assistant|><think>`.
- This fixes a train-inference mismatch where the model was trained to predict a token that is always provided at inference time, which is the likely root cause of repetitive thinking loops observed in SFT-tuned GLM-5.1 models.
- Updates 3 test assertions and docstring to reflect the new masking boundary. All 86 GLM5 renderer tests pass.

## Background

At inference time, the generation suffix for GLM-5.1 is `<|assistant|><think>` — the `<think>` token is always prepended to the model's generation. The V1 (`train-py`) renderer correctly masked this token via `commonprefix` logic. However, the V2 cookbook renderer was including `<think>` in the trained output, meaning the model was being trained to predict a token it would never need to generate. This one-token shift in the autoregressive training target likely caused the model to lose track of its generation start point, leading to repetitive reasoning loops.

### `training/renderer/glm5.py`
- For **terminal turns**: header is now `<|assistant|><think>`, output starts at reasoning content or `</think>`
- For **historical turns with preserved thinking**: header includes `<|assistant|><think>`, output starts at reasoning content
- For **historical turns with stripped thinking**: header remains `<|assistant|>`, output is `</think>` (unchanged)
- Updated docstring to document the header/output split explicitly

### `training/tests/unit/test_glm5_renderer.py`
- `test_weight_mask_only_covers_assistant_output`: asserts trained text starts with `</think>`, not `<think></think>`
- `test_weight_mask_with_thinking`: asserts `<think>` is not in trained text
- Tool call test with reasoning: asserts `<think>` is not in trained text

## Test plan
- [x] All 86 GLM5 renderer unit tests pass
- [ ] Re-run SFT training with fixed renderer to verify looping is resolved

### DIFF
--- a/training/renderer/glm5.py
+++ b/training/renderer/glm5.py
@@ -18,15 +18,19 @@ Role tag layout (as the shipped Jinja template emits them):
 - ``<|assistant|>...``     — see below
 - ``<|observation|>{content}`` — same
 
-Assistant turn layout:
+Assistant turn layout (full token sequence; header/output split noted):
 
 - **Terminal turn, ``enable_thinking=True`` (default), no reasoning content**::
 
       <|assistant|><think></think>{content}
 
+  Header (masked) = ``<|assistant|><think>``, Output (trained) = ``</think>{content}``
+
 - **Terminal turn, reasoning content provided**::
 
       <|assistant|><think>{reasoning}</think>{content}
+
+  Header (masked) = ``<|assistant|><think>``, Output (trained) = ``{reasoning}</think>{content}``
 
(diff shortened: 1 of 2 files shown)


Hecate0821 (approved):

morgendave:
Closing: PR #400 already covers the `<think>` masking fix at the `build_supervised_example` level. Our changes are redundant.

PR #404, opened 2026-04-29
Fix dataloader cursor accounting (RL, IGPO, DPO, SFT)

Hecate0821:
## Summary

Adds `RawRowCursor` (`training/utils/dataloader_cursor.py`): one shared cursor implementation for cumulative raw-source rows consumed (incl. drops / sample failures), used by all four training loops.

**RL / IGPO** (`rl_loop.py`, `igpo_loop.py`): each loop owns a `RawRowCursor(max_rows=len(all_rows))`. `cursor.resume(persisted, fallback=rollouts*prompts_per_step)` replaces the legacy step-derived formula; `cursor.record(raw_rows_from_stats(loop_stats, accepted_rows))` advances by raw rows so dynamic-filter drops and sample failures no longer drift the resume cursor; `ckpt.save(data_consumed=cursor.value)`.

**DPO** (`dpo_loop.py`): `_train_loop` takes `cursor: RawRowCursor` (caller pre-resumes from `resume_info.data_consumed`), drops the `prior_data_consumed` parameter and tuple return. Producer calls `cursor.record(delta)` per loader batch; multi-epoch replay calls `cursor.record(total_raw_rows)` per replay epoch. DCP and final saves persist `cursor.value`.

**SFT** (`sft_loop.py`): inline `raw_rows_consumed` int replaced with `cursor`. Epoch decomposition reads `cursor.value` instead.

**ORPO**: intentionally unchanged. Uses upfront-filter semantics (post-filter pairs in `pair_cache`); converting to raw-rows is a separate refactor.

`RLPromptDataset` reverts to its pre-PR shape (rows + `prompts_per_step` + `get_batch` + `__len__`); the resume / record_batch / data_consumed / rows_from_cursor methods that this PR previously added were extracted to `RawRowCursor`. `ResumeInfo.data_consumed` docstring (`checkpoints.py:63`) defines the contract: *cumulative raw rows from the source dataset (incl. drops / sample failures), across all runs*. `run_rl_loop` carries a TODO at the partial-batch dispatch site for a pre-existing tail-drop gap (drops/fails after a window's last full-batch dispatch are uncounted; not introduced here).

## References / Acknowledgment
- Same broad direction as AReaL's stateful dataloader recovery: https://github.com/inclusionAI/AReaL
- Follows slime's data-source cursor approach (`sample_offset` / data-source state) and its multimodal prompt-data handling: https://github.com/THUDM/slime

## Sensitive Data Check
- All commits scanned for credential patterns (API keys, gateway secrets, passwords, bearer tokens). No matches.
- Existing `FIREWORKS_API_KEY` references in `rl_loop.py` are unchanged environment-variable usage, not secret values.

## Tests

26 cursor tests pass locally (no `tinker` / GPU required):

```bash
pytest -q training/tests/test_dataloader_cursor.py \
            training/tests/test_igpo_dataloader_cursor.py
```

DPO unit tests in `training/tests/unit/test_dpo_loop.py` require `tinker` and run in CI:
- existing `_train_loop` tests updated for the new `cursor=` parameter
- `test_data_consumed_includes_render_drops` — cursor.value reflects raw rows incl. render drops
- `test_data_consumed_threads_prior_value_on_resume` — pre-resumed cursor accumulates new raw rows on top
- `test_data_consumed_grows_per_epoch_in_multi_epoch` — `cursor.record(total_raw_rows)` per replay epoch

Plus syntax + import sanity:
```bash
python -m py_compile training/utils/dataloader_cursor.py training/utils/data.py \
                     training/utils/rl/train.py training/utils/checkpoints.py \
                     training/recipes/rl_loop.py training/recipes/igpo_loop.py \
                     training/recipes/dpo_loop.py training/recipes/sft_loop.py
git diff --check
```

## Test Coverage Notes

- `test_dataloader_cursor.py` (26): `RawRowCursor` mechanics (resume clamping, fallback ambiguity at `(0, 0)` vs `(0, N)`, record accumulation/clamping/defensive paths), `raw_rows_from_stats` helper, end-to-end `run_rl_loop` integration with drops/fails, legacy-ckpt resume, tail-drops gap regression guard.
- `test_igpo_dataloader_cursor.py` (3): source-string regression guards on `igpo_loop.py` (cursor methods present, legacy step-derived formula absent).
- DPO `_train_loop` tests in `test_dpo_loop.py` cover render-drop accounting, prior-value resume, and multi-epoch replay growth.


xiaoyifan:
it seems `make_render_dataloader(... shuffle=False)` is not skipping the consumed rows yet in DPO?

xiaoyifan (approved):

xiaoyifan:
but having`better cursor.record(raw_batch_size)` before flattening is already way better

PR #408, opened 2026-04-30
renderer: drop incorrect minimax_m2 xfail; add renderer README

Hecate0821:
## Summary

- Drop the incorrect xfail marker on the \`minimax_m2-single-turn\` HF parity case. \`pytest.xfail()\` short-circuits before running the comparison, so the previous marker was applied based on the live-gateway divergence without ever verifying CPU divergence. Running the parity check directly shows the renderer matches HF byte-for-byte; the gateway is the odd one out (`<think>` vs HF/renderer's `<think>\n`). The right action is to keep the renderer as-is and let the test enforce parity going forward.
- Add a compact \`training/renderer/README.md\` that routes agents to \`skills/renderer/SKILL.md\` (implement) and \`skills/verifier/SKILL.md\` (validate) and embeds the verifier token-stream screenshot.
- Drop the now-redundant "Renderer skills" section from the top-level cookbook README; the renderer README owns that entry point.

## Test plan

- [x] \`pytest -q training/tests/unit/test_renderer_hf_parity.py\` → 7/7 pass
- [ ] CI \`Training CI / Unit And Import Tests\` job goes green on this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,21 +28,10 @@ training/           Training SDK recipes, utilities, and examples
   recipes/          Fork-and-customize training loop scripts
   utils/            Shared config, data loading, losses, metrics
   examples/         Worked examples (RL, SFT, DPO, ORPO)
-  verifier/         Renderer correctness validator + live React viewer
+  renderer/         Cookbook-local renderers + verifier (see training/renderer/README.md)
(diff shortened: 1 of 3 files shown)


Hecate0821:

PR #409, opened 2026-04-30
verifier: side-by-side renderer↔gateway prompt comparison

Hecate0821:
## Summary

When the renderer's prompt diverges from the live Fireworks gateway's tokenisation — the load-bearing condition for SFT correctness on Fireworks — the GUI now shows the two prompts in adjacent columns with the divergent suffix highlighted. This makes the training-inference distribution shift visible inside the verifier instead of buried in JSON.

- `probe.py` — persists the gateway's prompt tokens (+ decoded form) under `artifact.api.prompt.{tokens,decoded}`. Previously computed and discarded.
- `viewer/index.html` — new `PromptComparison` component, slotted above the token stream in each case. Only renders when `sanity.renderer_prompt_matches_api_prompt` is false AND the artifact has the new block. Older artifacts skip cleanly.

## End-to-end demo (kimi_k25_disable_thinking)

```bash
export FIREWORKS_API_KEY=fw_...
./run.sh
```

In the GUI:
- Renderer: `kimi_k25_disable_thinking` (tokenizer auto-fills to `moonshotai/Kimi-K2.5`)
- Model: `accounts/fireworks/models/kimi-k2p5`
- Type any prompt → `Run probe`

The amber `Prompt comparison` panel appears above the token stream, showing renderer's `<think></think>\n\n` block (matching HF) vs gateway's open `<think>\n` (the gateway doesn't honour `enable_thinking=false` for this serverless model). Divergent positions are red.

## Test plan

- [x] `pytest -q training/tests/unit/test_verifier_probe.py` (3 passed; artifact now has `api.prompt`)
- [ ] Live smoke: run a kimi_k25_disable_thinking probe via `./run.sh`, confirm the comparison panel appears with the structural divergence highlighted
- [ ] Live smoke: run a glm5 probe (renderer matches gateway), confirm the panel doesn't appear

### DIFF
--- a/training/renderer/verifier/utils/probe.py
+++ b/training/renderer/verifier/utils/probe.py
@@ -605,6 +605,12 @@ def run_probe(
                 "decoded": [_decode_one(tokenizer, t) for t in full_tokens],
             },
         },
+        "api": {
+            "prompt": {
+                "tokens": api_prompt_tokens,
+                "decoded": [_decode_one(tokenizer, t) for t in api_prompt_tokens],
+            },
+        },
         "completion": {
             "text": completion_text,
             "tokens": completion_tokens,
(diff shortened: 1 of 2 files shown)


Hecate0821:

PR #411, opened 2026-05-01
Validate unused reference infra config

Hecate0821:
## Summary
- Fail fast when a reference job ID is supplied while `needs_reference=False`.
- Fail fast when a reference training shape is supplied while `needs_reference=False`.
- Add unit coverage for both validation paths.

## Test Plan
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest training/tests/unit/test_infra_setup.py -q`
- `python -m compileall -q training/utils/infra.py training/tests/unit/test_infra_setup.py`
- `git diff --check`


Hecate0821:

PR #420, opened 2026-05-04
feat(deepseek_v4): add DeepseekV4Renderer with parity oracle and weight tests

lequytra:
## Summary

Port the upstream DSv4 encoder (`deepseek-ai/DeepSeek-V4-Flash/encoding/encoding_dsv4.py`) into the `tinker_cookbook` renderer abstraction so SFT/RLOR data renders byte-for-byte the same way DSv4 inference does.

DSv4 ships a hand-rolled Python encoder (no Jinja `chat_template`), so the encoder itself is the authoritative source of truth. We vendor it verbatim under `training/tests/_encoding_dsv4_oracle.py` (with a refresh recipe in the file header) and pin the renderer to it.

## What it covers

- `thinking` and `chat` modes
- `strip_thinking_from_history` flag, including the encoder's auto-disable when any message carries `tools`
- DSML-formatted parallel tool calls (`string="true"`/`"false"` arg branching)
- Tool-result reordering to match preceding `tool_calls` order (mirrors `sort_tool_results_by_call_order`)
- `tools` and `response_format` rendered into the system prompt
- Generation prompt: `<｜Assistant｜><think>` (thinking) / `<｜Assistant｜></think>` (chat)
- `parse_response` recovers `reasoning_content` + `content` + `tool_calls`

**Out of scope** (intentionally — call sites needing these should keep using the upstream encoder directly): `developer` role, `latest_reminder` role, internal task tokens (`action`/`query`/`authority`/`domain`/`title`/`read_url`), `wo_eos`, `reasoning_effort` preamble, `context` parameter.

## Code overview

```mermaid
flowchart LR
    subgraph caller["Caller (SFT loop / RL loop)"]
        msgs["List[Message]"]
    end

    subgraph renderer["DeepseekV4Renderer"]
        prep["_preprocess<br/>merge_tool→sort→drop_history"]
        rm["render_message<br/>system / user / assistant"]
        gs["_get_generation_suffix"]
        pr["parse_response"]
    end

    subgraph oracle["_encoding_dsv4_oracle (vendored, tests-only)"]
        enc["encode_messages"]
        parse["parse_message_from_completion_text"]
    end

    msgs -->|build_supervised_example| prep --> rm
    msgs -->|build_generation_prompt| prep --> gs
    rm -->|"(header, output, EOS)"| sup["ModelInput + weights"]
    gs -->|prompt prefix| gen["ModelInput"]

    sup -. "test parity" .-> enc
    gen -. "test parity" .-> enc
    sup -. "trained-span roundtrip" .-> parse
    pr -. "tool-call shape parity" .-> parse
```

## Testing strategy (89 tests, all passing)

Three layers of correctness checks:

1. **Format parity** — string and token IDs match `encoding_dsv4.encode_messages` across thinking-strip / thinking-keep / chat modes × 7 fixtures, plus tool-call argument formatting (string vs JSON-encoded mixed types) and parallel tool-result reordering.
2. **Train-inference prefix invariant** — the bedrock weight-mask check. For every supervised fixture, `build_generation_prompt(messages[:-1])` is a strict zero-weight prefix of `build_supervised_example(messages)`, and every token after the prefix has weight 1. Parametrized across all three modes × 7 fixtures.
3. **Trained-span roundtrip via the oracle parser** — decode the trained span and feed it to `encoding_dsv4.parse_message_from_completion_text`; it must reconstruct the original assistant message (reasoning + content + tool calls). Uniquely possible for DSv4 because the oracle ships a parser as well.

Plus targeted weight-mask tests:
- EOS trained as the assistant stop signal
- `<｜Assistant｜>` and the boundary `<think>` / `</think>` token masked (model never produces the framework-injected prefix)
- User / system / `<tool_result>` spans masked
- DSML tool-call body trained (`<｜DSML｜invoke>...<｜DSML｜parameter>` and arg values)

## Run locally

```bash
cd training
PYTHONPATH=../.. python -m pytest tests/unit/test_deepseek_v4_renderer.py -v
```

## Test plan

- [x] All 89 new tests pass against `deepseek-ai/DeepSeek-V4-Flash` from HF Hub
- [x] Existing renderer tests (GLM-5, MiniMax-M2, Nemotron, Gemma-4) still pass — 144 / 144 (32 skipped on tokenizer availability, unchanged from main)
- [x] `black` clean
- [x] No lint errors

### DIFF
--- a/training/renderer/__init__.py
+++ b/training/renderer/__init__.py
@@ -10,6 +10,7 @@ Importing this package eagerly imports every contained renderer module so
 the registrations take effect.
 """
 
+from training.renderer import deepseek_v4 as _deepseek_v4  # noqa: F401  (registers "deepseek_v4")
 from training.renderer import gemma4 as _gemma4  # noqa: F401  (registers "gemma4")
 from training.renderer import glm5 as _glm5  # noqa: F401  (registers "glm5")
 from training.renderer import minimax_m2 as _minimax_m2  # noqa: F401  (registers "minimax_m2")
--- a/training/renderer/deepseek_v4.py
+++ b/training/renderer/deepseek_v4.py
@@ -1,0 +1,704 @@
+"""Renderer for DeepSeek-V4-Flash chat templates.
+
+This is a port of the upstream Python encoder shipped at
+``deepseek-ai/DeepSeek-V4-Flash/encoding/encoding_dsv4.py`` (license: MIT)
+into the ``tinker_cookbook`` renderer abstraction.
+
+DeepSeek-V4 uses a hand-rolled DSML-flavored format rather than a Jinja
+chat template:
+
+- ``<｜begin▁of▁sentence｜>`` once at the very start of the conversation.
+- System messages emit their content with no role tag.
+- User messages emit ``<｜User｜>{content}`` (or, after merging, a
+  ``\\n\\n``-joined sequence of text blocks and
+  ``<tool_result>...</tool_result>`` blocks for prior tool responses).
+- Assistant messages emit ``{thinking_block}{content}{tool_calls}<EOS>``.
+  Thinking is a literal ``<think>{reasoning}</think>`` and is always
+  preceded by a ``<｜Assistant｜>`` role tag emitted by the *boundary*
+  between the previous user-or-developer message and this assistant turn.
+
+The boundary suffix has three forms:
+
+* ``<｜Assistant｜><think>``  — thinking-mode terminal turn, OR thinking
+  mode without history-stripping.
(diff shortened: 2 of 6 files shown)


xiaoyifan:
This will  break multi-turn case. check https://github.com/fw-ai/cookbook/pull/418, it's better to have:
```
class DeepseekV4Renderer(Renderer, DisaggregateMultiTurnMixin)
```

so it has the build_supervised_examples.


but you can also check it in if there's no other blocker, i can rebase and add it for you

xiaoyifan (approved):

PR #426, opened 2026-05-06
training: pre-validate promote_checkpoint and document env vars

Hecate0821:
## Summary

Two small alignment fixes against the public Training-API surface:

- **`examples/tools/promote_checkpoint.py`** now calls `validate_output_model_id(...)` before `promote_checkpoint(...)` and exits with the validator errors on a bad ID. A server-side rejection after the sampler blob has been staged orphans the blob, so the example should demonstrate the safe flow.
- **`training/.env.example`** is new — a single discoverable file listing every variable the cookbook reads (`FIREWORKS_API_KEY`, optional `FIREWORKS_BASE_URL`, optional `WANDB_API_KEY` / `WANDB_MODE`). `training/README.md` now points users at it.

No public API changes, no recipe behavior changes.

## Test plan

- [ ] `python training/examples/tools/promote_checkpoint.py --job-id <real-job> --base-model accounts/fireworks/models/qwen3-8b --output-model-id INVALID-UPPERCASE` exits 2 with validator errors and does not call the promote API.
- [ ] Same command with a valid output model ID promotes successfully (regression check — happy path unchanged).
- [ ] `cp training/.env.example training/.env` and a recipe run still picks up `FIREWORKS_API_KEY` via `python-dotenv`.

### DIFF
--- a/training/.env.example
+++ b/training/.env.example
@@ -1,0 +1,17 @@
+# Copy to `.env` (same directory) and fill in real values.
+# `python-dotenv` loads this automatically when recipes/examples call
+# `load_dotenv()` from within the cookbook tree.
+
+# --- Fireworks ---
+# Required. Get one at https://fireworks.ai/account/api-keys
+FIREWORKS_API_KEY=
+
+# Optional. Override the control-plane host. Default: https://api.fireworks.ai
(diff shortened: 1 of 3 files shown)


Hecate0821:
Closing — not needed for this alignment pass.

PR #443, opened 2026-05-16
perf(sft): drop multi-turn renderer splits that have zero training weight

benjibc:
## Summary

`_render_one_worker` already filters rendered datums by token length
(`2 <= len <= max_seq_len`). It does **not** filter out datums whose
`token_weights` are all zero. On datasets with per-message weights
(SFT teacher-force, mixed-supervision conversations, RLOR-style teacher
mimicking), the multi-turn renderer split emits examples whose target
assistant turn has `weight=0` — those examples are pure context for
the prefix and produce zero gradient, but still cost a full forward
pass.

## Measured impact

On one such dataset (Kimi K2.5/K2.6 chat-with-tools, mixed weights):

| Stat | Value |
|---|---|
| Assistant turns with `weight=0` | **73.5 %** |
| Rows with >1 user turn (renderer splits) | 66.1 % |
| Renderer-emitted splits that target a `weight=0` assistant | **88.1 %** |
| Per-step tokens shipped to the trainer that produced zero loss | **~73 %** |

## Fix

A single condition added to the existing datum filter:

```python
datums = [
    rendered.datum
    for rendered in rendered_examples
    if 2 <= len(rendered.token_ids) <= _worker_state["max_seq_len"]
    and any(w != 0 for w in rendered.token_weights)   # NEW
]
```

If a rendered example has zero training weight across all positions,
drop it before the trainer ever sees it.

## Notes

- Same code path is dispatched by managed SFT V2
  (`training_orchestrator._run_sft` → `training.recipes.sft_loop.main`),
  so the perf win flows through to every customer SFT job automatically.
- A more thorough fix would prune zero-loss user-turn boundaries inside
  `tinker_cookbook`'s `build_supervised_examples` so they're never
  rendered in the first place (saves CPU on the dataloader side too).
  This cookbook-side filter is the smallest diff that captures the
  GPU-side win.

---

> [!NOTE]
> **Medium Risk**
> Changes SFT dataset filtering to skip examples with zero training weight, which can alter effective training data/step counts on weighted datasets if weights are mis-specified or all-zero.
> 
> **Overview**
> **SFT rendering now drops zero-loss examples.** `_render_one_worker` filters out renderer-emitted datums whose `token_weights` are entirely non-positive, in addition to the existing sequence-length bounds, to avoid wasting forward passes on examples that contribute no gradient.
> 
> Unit tests update their `render_messages_to_datums` stubs to include `token_weights` so the new filter path is exercised and existing training-loop behaviors remain covered.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7029864e2b67f20e26957c091fe760eebbcb8ae9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/training/recipes/sft_loop.py
+++ b/training/recipes/sft_loop.py
@@ -114,7 +114,7 @@ def _render_one_worker(row: dict) -> tinker.Datum | list[tinker.Datum] | None:
         rendered.datum
         for rendered in rendered_examples
         if 2 <= len(rendered.token_ids) <= _worker_state["max_seq_len"]
-        and any(w != 0 for w in rendered.token_weights)
+        and any(w > 0 for w in rendered.token_weights)
     ]
     if not datums:
         return None
--- a/training/recipes/sft_loop.py
+++ b/training/recipes/sft_loop.py
@@ -114,6 +114,7 @@ def _render_one_worker(row: dict) -> tinker.Datum | list[tinker.Datum] | None:
         rendered.datum
         for rendered in rendered_examples
         if 2 <= len(rendered.token_ids) <= _worker_state["max_seq_len"]
+        and any(w != 0 for w in rendered.token_weights)
     ]
     if not datums:
         return None
--- a/training/tests/unit/test_sft_loop.py
+++ b/training/tests/unit/test_sft_loop.py
@@ -200,7 +200,7 @@ def test_main_raises_when_all_examples_are_filtered(tmp_path, monkeypatch):
     monkeypatch.setattr(
         module,
         "render_messages_to_datums",
-        lambda *args, **kwargs: SimpleNamespace(token_ids=[1], datum={"id": "too-short"}),
+        lambda *args, **kwargs: SimpleNamespace(token_ids=[1], token_weights=[1.0], datum={"id": "too-short"}),
     )
     monkeypatch.setattr(module, "ReconnectableClient", FakeClient)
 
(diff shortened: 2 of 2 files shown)


websterbei (approved):

training/recipes/sft_loop.py
```diff
@@ -114,6 +114,7 @@ def _render_one_worker(row: dict) -> tinker.Datum | list[tinker.Datum] | None:
         rendered.datum
         for rendered in rendered_examples
         if 2 <= len(rendered.token_ids) <= _worker_state["max_seq_len"]
+        and any(w != 0 for w in rendered.token_weights)
```
websterbei: it could be float

websterbei: maybe safer using >= 0

websterbei (approved):

benjibc:
Updated to use `w > 0` (`292f0c3`).

Reasoning: weights here are floats. `w > 0` is semantically clearer ("has positive training signal") and defensive against any potential negative-weight or tiny-float edge case while keeping the behavior on `weight == 0` datums (drop) unchanged. `>= 0` would also keep `weight == 0` and defeat the filter, so I read your comment as "switch from `!= 0` to a positive-signal check" — let me know if you meant something different.

training/recipes/sft_loop.py
websterbei: ^ the first comment is not true. I'm forgetting Python syntax

xiaoyifan (approved):

PR #614, opened 2026-07-26
NEW case study: embedding fine-tuning case study: customer support search

sinanuozdemir:
Contrastive fine-tuning of Qwen3-Embedding-8B on Airbnb Help Center scenarios, using the existing training/recipes/embedding_loop.py recipe. Retrieving the article that *governs* a situation is a domain-defined relation the base model has no way to learn from pretraining, and the base numbers showed the headroom: Recall@100 0.991 against MRR 0.525, meaning it found the right neighborhood and buried the answer at rank 2.

72 steps over 769 pairs lifts held-out MRR 0.525 -> 0.707 (+35%) and nDCG@10 0.602 -> 0.771, moving the correct article from rank ~1.9 to ~1.4. Gains and losses are asymmetric: the biggest wins recover queries from ranks 60-140, while the worst regression is rank 2 -> 13.

Adds retrieval_eval.py, the retrieval harness the cookbook was missing: batched /v1/embeddings, cosine top-k, and nDCG/Recall/MRR/MAP via pytrec_eval.

Four platform behaviours are documented because each silently produces wrong results or a confusing failure:
- the query instruction is asymmetric and baked into the weights, so eval must reproduce the training prefix exactly
- the trainer needs a tunable -ft-base model plus a training shape; the serving id is rejected as EMBEDDING_MODEL
- a promoted checkpoint must be re-registered with kind=EMBEDDING_MODEL and base_model_details (both create-time only) and deployed with an embedding deployment shape, or embeddings serve on the generative path
- dedicated deployments require model#deployment routing

The notebook asserts input-form invariance before reporting any metric so a misrouted deployment fails loudly instead of returning plausible but wrong vectors.


terryfireworks (approved):
looks good.

PR #619, opened 2026-07-27
Remove deprecated trainer accelerator configuration

Hecate0821:
## Summary

- remove `accelerator_type` and `accelerator_count` from cookbook trainer configuration, including the legacy `InfraConfig` surface
- stop forwarding those fields through the cookbook-to-SDK service adapter and remove them from sync/async RL result payloads
- reject the removed fields with an actionable error from direct Python construction, YAML provisioning, and Hydra CLI overrides
- remove stale trainer accelerator flags/fixtures from cookbook E2E coverage and update the canonical training skill guidance

## Why

Trainer accelerator selection is shape-owned and cookbook clients can no longer set accelerator type or count. Keeping deprecated fields as ignored inputs made configurations look valid while doing nothing. This change intentionally drops compatibility: callers must remove those fields and use `training_shape_id` when they need an explicit trainer shape.

No SDK changes are included.

## User impact

Passing either removed field now fails immediately with:

> Cookbook clients cannot set trainer accelerator type or count; select a training shape with `training_shape_id` instead.

The fields are also absent from dataclass metadata and constructor signatures, so generated/config-driven clients no longer advertise them.

## Validation

- `uv run --project training pytest -q training/tests/unit training/provision/tests`
  - 3,191 passed, 217 skipped, 79 xfailed
- protected path regression suite:
  - `training/tests/unit/test_serverless.py`
  - `training/tests/unit/test_sft_loop.py`
  - `training/tests/unit/test_dpo_loop.py`
  - 79 passed
- `uv run --project training pytest -q training/tests/test_smoke_imports.py`
  - 68 passed
- `training/.venv/bin/python skills/validate_skills.py`
- Ruff lint on every changed Python file
- `git diff --check`
- `bash -n training/tests/e2e/run_frozen_lake_b300.sh`

Remote E2E jobs were not launched because they provision live/billable training resources.

## Retrospective / scope audit

Every change is needed for one of these contracts:

- **Config surface:** `TrainerConfig` and legacy `InfraConfig` must stop advertising the fields while still producing the required migration error.
- **Provisioning boundary:** the YAML/Hydra loader previously ignored unknown dataclass keys, so it needs an explicit rejection before filtering; the shared service adapter must stop passing the removed values to the SDK.
- **Recipe output:** sync and async RL carried newly-added accelerator result fields plus lookup code solely for those fields, so both are removed consistently.
- **Cookbook callers:** E2E fixtures and the FrozenLake smoke script still set trainer accelerators; removing them prevents the cookbook's own examples/tests from teaching or exercising an unsupported path.
- **Skill contract:** repository guidance requires compatibility changes to update the canonical training skill, so stale deprecated/ignored wording is replaced with the new hard-error contract.
- **Regression coverage:** direct construction, dataclass discovery, static YAML, Hydra CLI overrides, RL results, serverless SFT, and managed SFT/DPO are all covered.

Intentionally unchanged:

- the Fireworks SDK
- deployment-specific accelerator controls, which are a separate supported deployment contract
- server-resolved accelerator reads used by managed SFT/DPO runner metadata and provisioning progress; these are reporting/billing inputs, not client configuration
- the serverless SFT setup and `mark_serverless` billing path

### DIFF
--- a/skills/fireworks-training/references/rl-custom-loss.md
+++ b/skills/fireworks-training/references/rl-custom-loss.md
@@ -134,8 +134,10 @@ All live on `rl_loop.Config`:
 | `router_replay`, `router_replay_completion_only` | `True`, `True` | Replay MoE routes for generated tokens by default. Set completion-only to `False` only when full-sequence replay is worth the serving cost of `echo=True`. |
 | `grad_accumulation_normalization` | `None` | No server-side normalization by default. Use `NUM_LOSS_TOKENS` for raw-sum losses. See [`rl-gradient-accumulation.md`](rl-gradient-accumulation.md). |
 
-Shape-owned fields (`accelerator_type`, `node_count`, and `custom_image_tag`)
-come from the training profile; never hand-set them.
+Trainer accelerator type and count are not cookbook config fields; select them
+indirectly with `training_shape_id`. Other shape-owned fields such as
+`node_count` and `custom_image_tag` come from the training profile; never
+hand-set them.
 
 ## Do not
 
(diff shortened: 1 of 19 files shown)


Hecate0821:
Superseded by the mergeable source-of-truth PR: https://github.com/fw-ai/fireworks/pull/37977. This repository accepts only generated promotion PRs, so the cookbook change is now scoped to `public-repos/cookbook/**` in the source repository and will arrive here through the normal promotion workflow.

PR #621, opened 2026-07-28
Add reproducible GLM-5.2 Terminal-Bench RL example

renfeichen-fw:
## Summary

- add the proven GLM-5.2 full-parameter OpenCode Terminal-Bench recipe as a
  cookbook example
- preserve `train.py` and `train_fireworks_glm5p2.sh` byte-for-byte from
  rllm-org/rllm#779 at `5f2fed491589790457c6bb292734415ea108463a`
- bundle the approved fixed 48-task curriculum, including its original order,
  task environments, solutions, and verifiers
- add deterministic train/eval registration, pin Terminal-Bench 2.1 to
  `terminal-bench/terminal-bench-2-1@6`, and reject train/eval leakage
- document the exact 12-step, four-epoch configuration and link the reference
  W&B run

## Why

The existing experiment was numerically validated, but its exact recipe and
training subset only lived in the rLLM development checkout. This makes the
experiment reproducible from the Fireworks training cookbook without
translating the loop or changing its training semantics.

## Validation

- dependency resolution: `uv pip install --dry-run --overrides ... -e './training[terminal-bench]'`
- dataset registration: 48 ordered train tasks and 89 pinned eval tasks with
  zero overlap
- source parity: recipe and launcher SHA-256 match the pinned rLLM revision
- data parity: all 845 non-cache task files match the approved source bytes
- `pytest -q training/tests/unit/test_terminal_bench_glm5p2.py` (`2 passed`)
- Ruff check/format, `bash -n`, and `git diff --check`

No trainer, deployment, paid inference, or GPU workload was launched.


renfeichen-fw:
Superseded by fw-ai/fireworks#38072. Cookbook content must originate from FireTitan staging; the normal promotion workflow will open the bot-authored public PR after the staging change merges.

PR #629, opened 2026-08-03
fix(training): clean up dedicated recipe trainers on close

lequytra:
## Summary

- enable `cleanup_trainer_on_close` for dedicated SFT, ORPO, and embedding recipes
- retain the existing `ExitStack` lifecycle so cleanup runs on normal completion and exceptions
- add call-site regression tests for all three changed recipes
- verify DPO continues to request the same cleanup policy

## Why

These recipes registered `service.close()` but inherited `build_service_client(..., cleanup_trainer_on_close=False)`. Closing therefore left SDK-created trainers running until inactivity cleanup, even though the recipes previously cleaned up newly created trainers and the cleanup documentation requires this flag.

The SDK applies trainer cleanup only when it created the trainer, so reattached trainers remain preserved.

## Testing

- explicit cleanup-policy assertions for SFT, DPO, ORPO, and embedding: **4 passed**
- `uvx ruff check` on all changed files
- full SFT, ORPO, and embedding unit modules: **50 passed**
- complete `training/tests/unit` run: **2,987 passed, 358 skipped, 75 xfailed, 26 failed**; the failures are unrelated environment/version mismatches (Tinker 0.22.2 vs tests requiring 0.23.0, stale SDK schema, and existing DeepSeek renderer expectations)

### DIFF
--- a/training/recipes/embedding_loop.py
+++ b/training/recipes/embedding_loop.py
@@ -338,6 +338,7 @@ def main(config: Config):
             max_context_length=max_context_length,
             learning_rate=cfg.learning_rate,
             trainer=cfg.trainer,
+            cleanup_trainer_on_close=True,
         )
         stack.callback(service.close)
         training_client = service.create_training_client(cfg.base_model, lora_rank=cfg.lora_rank)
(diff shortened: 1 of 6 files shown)


lequytra:
Superseded by the authoritative source PR fw-ai/fireworks#39534. Cookbook main only accepts generated promotion PRs, so this fix will be promoted from public-repos/cookbook after the source PR merges.

PR #633, opened 2026-08-04
Fix DeepSeek V4 multi-turn SFT rendering

xiaoyifan:
## Summary

- make `DeepseekV4Renderer` use the shared multi-turn disaggregation path
- split `ALL_ASSISTANT_MESSAGES` rows into one supervised datum per user turn when historical thinking is stripped
- keep non-training modes on the existing single-datum fast path
- add production-path token and loss-mask parity coverage against DeepSeek's official `encoding_dsv4.py`
- prevent the DeepSeek V4 tokenizer suite from silently skipping when `AutoTokenizer` cannot load the unregistered `deepseek_v4` model config

## Root cause

The default DeepSeek V4 renderer strips thinking from historical assistant turns, so it does not have the sequence extension property. Multi-turn SFT therefore needs `build_supervised_examples` to render each user-turn prefix independently.

The renderer previously inherited the base implementation, which raised `NotImplementedError` when production attempted to render a multi-turn `ALL_ASSISTANT_MESSAGES` row.

## Behavior

For a two-turn conversation, training now produces:

- datum 1: first assistant reasoning/answer is the target
- datum 2: the first assistant is masked history and the second assistant reasoning/answer is the target

For each datum, the test independently derives:

1. expected token IDs from the official DeepSeek V4 encoder plus HF tokenizer
2. expected mask boundaries from the official generation prompt versus full supervised sequence

Both token IDs and masks must match the production `render_messages_to_datums` output exactly.

## Validation

- `94 passed` — DeepSeek V4 renderer suite
- `110 passed` — shared disaggregation matrix
- `67 passed` — production supervised rendering suite
- `1892 passed, 82 skipped, 78 xfailed` — renderer property suite
- `ruff check`
- `git diff --check`

A managed GPU training canary is still required after the branch/image is deployed; the existing scheduled training smoke currently covers Qwen rather than DeepSeek V4.

### DIFF
--- a/training/renderer/deepseek_v4.py
+++ b/training/renderer/deepseek_v4.py
@@ -87,6 +87,8 @@ from tinker_cookbook.renderers.base import (
 )
 from tinker_cookbook.tokenizer_utils import Tokenizer
 
+from training.renderer._disaggregate_mixin import DisaggregateMultiTurnMixin
+
 # ── Special tokens (must match encoding_dsv4.py exactly) ────────────────────
 # NOTE: ``｜`` is U+FF5C FULLWIDTH VERTICAL LINE (not ASCII ``|``).
 #       ``▁`` is U+2581 LOWER ONE EIGHTH BLOCK (not ASCII ``_``).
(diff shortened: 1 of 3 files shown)


xiaoyifan:
Superseded by fw-ai/fireworks#39765. Cookbook changes must land through the reviewed `public-repos/cookbook` staging surface and the promotion bot; this direct public-repo PR cannot pass the promotion source gate.

PR #650, opened 2026-08-10
docs(skills): read training shapes via the version collection, not the parent

charliegu-fw:
## Why

A customer (`dev-a09gkpuhep24`, Pylon 61134) reported `rpc error: code = PermissionDenied` from `firectl training-shape list` and `firectl training-shape get accounts/fireworks/trainingShapes/qwen3-30b-a3b-instruct-2507-128k`, and read it as proof that their Dedicated Training API entitlement had not been applied. It had been applied — the commands simply are not readable by customer principals, and are not an entitlement signal.

Our own skill docs sent support down that path: `training-api.md` recommended `firectl training-shape list` as the dedicated-work **access preflight**, and `sdk-shapes.md` recommended it as the way to see "the full list of shapes in your account".

## Root cause

Verified against the prod control plane (`api.fireworks.ai`, read-only GETs):

- Public visibility is modelled on **`TrainingShapeVersion.public`**. The parent `TrainingShape` resource has no such field, so it falls back to plain account ownership — reading `accounts/fireworks/trainingShapes/<shape>` is a cross-account read and is denied for every customer principal.
- `ListTrainingShapes` rejects a wildcard parent (`wildcards cannot be used in parent for ListTrainingShapes`), so `training-shape list` can only ever list one account's shapes. `accounts/dev-a09gkpuhep24/trainingShapes` has `totalSize: 0`, so the command could never surface the shared catalog even if it were permitted.
- `ListTrainingShapeVersions` *does* accept the wildcard parent, which is why `FireworksClient.resolve_training_profile()` reads `/versions?filter=latest_validated=true` and never touches the parent resource.
- The shape in question is healthy and public: version `gwy64vxm`, `public: true`, `POLICY_TRAINER`, 8x `NVIDIA_B200_180GB`, 128k context.

The asymmetry is visible in the published surface too: the REST API reference documents Get/List for deployment shapes and their versions, but exposes no training-shape endpoint at all, and the training-shapes page tells users to read fields with `training-shape-version get`.

## Changes

Docs only, no code.

- `references/sdk-shapes.md` — replace the shape-listing snippet with `training-shape-version list/get`, and add a "Why not `firectl training-shape list` / `get`" section explaining the parent-vs-version split.
- `references/training-api.md` — access preflight now uses `training-shape-version list --base-model <model>`, with an explicit warning not to use the parent commands as a preflight.
- `references/error-reference.md` — new row for the `PermissionDenied` symptom, stating it is expected and not an entitlement signal.
- `SKILL.md` — degraded-source fallback points at the version collection.

## Testing

Docs-only change; `training/tests/unit/test_llms_txt.py` covers `llms.txt`, which is untouched. Behavioural claims were verified with read-only GETs against the prod control plane as described above.

<div><a href="https://cursor.com/agents/bc-937d947c-d99b-4299-a86c-5de3b68bc3a0?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-937d947c-d99b-4299-a86c-5de3b68bc3a0&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Hecate0821:
Please PR to staging repo

PR #652, opened 2026-08-10
Document GPU quota accounting and add an audit tool for dashboard mismatches

charliegu-fw:
## Why

A customer reported that `training-b200-count` showed **64 GPUs used** while they could only account for **32** on the fine-tuning dashboard, and asked how to find the missing 32 and who was using them. This was initially read as a possible platform bug.

It is not, mostly. The gap is an accounting model that is invisible from the fine-tuning page, and the skill had no guidance for reconciling the two. Investigating against live control-plane data also surfaced a real leak class the docs never mentioned.

## What the mismatch actually is

Three rules, each verified against live account data (three independent quota buckets reconcile to the GPU):

1. **An RL job is a trainer *plus* a rollout deployment, and both charge `training-<gpu>-count`.** The fine-tuning page (`?type=rlor-trainer`) lists the trainer half only, so counting jobs there undercounts a running fleet by roughly half. The per-replica trainer width also lives on the *training shape*, not the job, so a job that reads as "8 GPUs" can be far wider.
2. **Quota is charged on the autoscaling ceiling**, `acceleratorCount * maxReplicaCount`, not on running replicas. A cold or unschedulable deployment holds its full ceiling while serving nothing, and `--min-replica-count 0` alone does not release it.
3. **Trainer attachment decides the bucket.** A rollout attached to a live trainer charges `training-*`; a detached one charges `global--*`. So "ignore `global--*`, that's production inference" hides leaked *training* capacity.

## The leak worth knowing about

Deleting a trainer job tombstones it — `ListRlorTrainerJobs` filters `JOB_STATE_DELETED`, so it disappears from the dashboard immediately — but does **not** tear down the paired rollout deployment. Cookbook rollout deployments are pinned to `min_replica_count == max_replica_count` (`training/utils/config.py`), so the orphan can never scale itself to zero and no idle default rescues it. The recipes' `cleanup_on_exit` only runs on a clean client close, so a `SIGKILL`ed driver or a trainer deleted from the dashboard skips it entirely, and both cleanup paths only *warn* on failure.

The result holds GPUs indefinitely, is invisible on the fine-tuning page, and — because the trainer row is gone — has no signpost left.

## Changes

**New tool** — `training/examples/tools/audit_gpu_usage.py`, read-only (GETs only, prints the `firectl` commands for cleanup rather than running them). One pass over trainers, deployments, quotas, and shapes; charges each resource by the rules above, attributes it to an owner, and prints what fails to reconcile.

Two details that took work to get right:

- Deployments carry **no `createdBy`**, so per-user attribution routes through the trainer. For an orphan whose trainer was deleted, a direct GET still resolves the tombstoned row for 30 days and names whoever deleted it. Anything untraceable is reported as `unattributed` rather than assumed to be production inference.
- An orphan created by an older SDK has neither the `fireworks-training-sdk/managed-rollout` annotation nor a `<trainer-id>-rollout` name. Its **deployment shape** is the only remaining signal, so the tool derives the rollout-shape set from each training shape's `deploymentShapeVersion` instead of hard-coding a naming prefix. Without this a 16-GPU orphan reads as production inference.

**New reference** — `skills/fireworks-training/references/gpu-quota-accounting.md`: the accounting model, the audit procedure, attribution, and how to distinguish a stale control-plane reservation (not releasable by the account owner) from something a customer can clean up.

**Corrections to existing guidance:**

| File | Was |
|---|---|
| `getting-started.md` | Claimed quota "is not a reservation" and "you pay only for what runs". Billing follows GPU-seconds, but quota follows the ceiling. |
| `deploy-and-troubleshoot.md` | Described `scale_to_zero` without noting min-alone does not release quota, and that no idle default rescues a `min == max` deployment. Orphan-recovery section didn't mention rollout deployments at all. |
| `rl-hotload.md` | Documented trainer tombstoning without noting it leaves the rollout deployment running. |
| `error-reference.md` | No triage row for this symptom, so it routed toward escalation. |

## Testing

`training/tests/unit/test_audit_gpu_usage.py` — 12 tests over the pure accounting functions, with fixtures mirroring real control-plane payloads (a job pinning a shape *version*, an attached managed rollout, an annotation-less orphan, a plain inference deployment). Covers ceiling-vs-running charges, shape-version resolution, tombstoned jobs holding nothing, orphan classification with and without the shape catalogue, tombstone-based attribution, and quota usage that no resource explains.

```
12 passed
```

Run with `--noconftest`: the repo's unit `conftest.py` imports `torch` / `tinker_cookbook`, which are absent in this environment. That collection error is pre-existing and reproduces identically on existing test files (verified against `test_checkpoints.py`), so it is not introduced here.

`python skills/validate_skills.py` passes.

The tool was also run against the live account from the report; all quota buckets reconcile to 0 unexplained except one stale `training-b300-count` reservation with no B300 resource anywhere in the account, which is the escalation path the doc now describes.

<div><a href="https://cursor.com/agents/bc-7dd25b42-a0d2-43af-9aba-84ca2fc4290f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7dd25b42-a0d2-43af-9aba-84ca2fc4290f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Hecate0821:
Please PR to staging repo

PR #693, opened 2026-09-10
Require deployment shapes: block shapeless deployment creation

greg-fw:
## Why

Implements the cookbook side of the "Discourage / Disallow Shapeless Deployments" RFC. Shapeless deployments — creating a deployment without passing a shape, whether or not accelerator fields are set — skip validated configuration and are the most common cause of failed deployment creations on Fireworks. The cookbook should never create, allow, or recommend one.

Related: fw-ai/docs#1294 (customer-facing docs language), fw-ai/fireworks#47689 / #47682 / #47681 (docs updates), fw-ai/fireworks#48393 (firectl shape visibility).

### Block the shapeless path in code

- `DeployConfig.to_deployment_config` now raises `ValueError` when `deployment_shape` is unset. No production callers today (compat surface only), so this tightens the contract without changing behavior — a fork that strips the shape config now gets an error instead of a shapeless deployment.
- Removed the deprecated `DeployConfig.deployment_accelerator_type` field — it existed only for the legacy shapeless path and had no non-test callers. `skip_shape_validation` is hardcoded `False`.

### Close the silent shapeless hole in distillation (the one behavioral change)

- Auto-created teacher deployments now resolve a shape via per-teacher `TeacherConfig.deployment_shape` → run-level `teacher_deployment_shape` → the resolved **student deployment shape** (`service.deployment_shape`). Previously a run with no shape configured anywhere silently created a shapeless teacher deployment.
- Impact: distillation runs that configured no teacher shape now get the student's shape (potentially a different GPU class than the backend's shapeless default). Heterogeneous teachers (different model than the student) need an explicit shape — documented in the skills docs. Overlaps with #691 (lagged teacher OPD), which also touches teacher code.

### Docs / skills

- `training/README.md`: new "Do not create deployments without a shape" section — shapeless definition, canonical risk language, deprecation note, how the cookbook resolves shapes, and the override-vs-omit distinction (overriding fields on top of a shape is fine; the shape owns accelerator selection).
- `skills/fireworks-training/references/sdk-shapes.md`: deployment-shape section rewritten — shapes mandatory, `to_deployment_config` rejects shapeless, per-model shape discovery via firectl.
- `skills/fireworks-training/references/sdk-distillation.md`: documents the enforced teacher shape resolution order.

Language mirrors the canonical wording from fw-ai/docs#1294 and its review: "most common cause of failed deployment creations", "the unshaped path may be deprecated in the future", contact Fireworks if no shape fits.

## Deliberately different from the customer docs

docs#1294 keeps the manual path visible as a last resort for customers; this PR **rejects** it in cookbook code — the cookbook itself should never emit a shapeless create.

## Testing

- New + updated tests: `tests/unit/test_sdk_compat.py` (shapeless rejection, shape forwarding, removed-field check), `tests/unit/test_distillation.py` (student-shape default, runtime-level assertion on the created teacher `DeploymentConfig`).
- CI smoke set (`test_smoke_imports`, `test_client`, `test_sdk_compat`, `test_serverless`, `test_service`, `test_runner`, `test_dpo_loop`): **210 passed, 1 skipped**.
- Full unit suite: **5165 passed**. 5 failures in `test_renderer_multimodal_rl.py` / `test_rollout_assembler.py` and 4 collection errors (`eval_protocol` missing) are pre-existing on main — verified by stashing this diff and re-running on a clean tree.

### DIFF
--- a/skills/fireworks-training/references/sdk-distillation.md
+++ b/skills/fireworks-training/references/sdk-distillation.md
@@ -172,9 +172,10 @@ cfg.multi_teacher = MultiTeacherConfig(
   deployments. `teacher_deployment_shape` sets the run-level default; individual
   `TeacherConfig.deployment_shape` can override it. When neither is set,
   teachers default to the resolved student deployment shape — set one
-  explicitly for heterogeneous teachers on a different model than the student,
-  since a shapeless teacher deployment skips validated configuration and fails
-  far more often at creation.
+  explicitly for heterogeneous teachers on a different model than the student.
+  Do not create teacher deployments without a shape: shapeless deployments are
+  the most common cause of failed deployment creations, and the unshaped path
+  may be deprecated in the future.
 - Per-teacher metrics use `teacher_route/<slug>/scored` and
   `teacher_route/<slug>/inflight`; skewed datasets can leave some teacher
   deployments underused.
--- a/skills/fireworks-training/references/sdk-distillation.md
+++ b/skills/fireworks-training/references/sdk-distillation.md
@@ -170,7 +170,11 @@ cfg.multi_teacher = MultiTeacherConfig(
   deployments.
 - `teacher_replica_count` controls replicas for auto-created frozen teacher
   deployments. `teacher_deployment_shape` sets the run-level default; individual
-  `TeacherConfig.deployment_shape` can override it.
+  `TeacherConfig.deployment_shape` can override it. When neither is set,
+  teachers default to the resolved student deployment shape — set one
+  explicitly for heterogeneous teachers on a different model than the student,
+  since a shapeless teacher deployment skips validated configuration and fails
+  far more often at creation.
 - Per-teacher metrics use `teacher_route/<slug>/scored` and
   `teacher_route/<slug>/inflight`; skewed datasets can leave some teacher
   deployments underused.
--- a/skills/fireworks-training/references/sdk-distillation.md
+++ b/skills/fireworks-training/references/sdk-distillation.md
@@ -169,13 +169,17 @@ cfg.multi_teacher = MultiTeacherConfig(
   region used by the SDK-managed student sampler and auto-created teacher
   deployments.
 - `teacher_replica_count` controls replicas for auto-created frozen teacher
-  deployments. `teacher_deployment_shape` sets the run-level default; individual
-  `TeacherConfig.deployment_shape` can override it. When neither is set,
-  teachers default to the resolved student deployment shape — set one
-  explicitly for heterogeneous teachers on a different model than the student.
-  Do not create teacher deployments without a shape: shapeless deployments are
-  the most common cause of failed deployment creations, and the unshaped path
-  may be deprecated in the future.
+  deployments. Shape resolution for auto-created teacher deployments (this is
+  enforced by the recipe code in `training/recipes/distillation_loop.py`, not
+  a user action): per-teacher `TeacherConfig.deployment_shape` wins, then the
+  run-level `teacher_deployment_shape`, and when neither is set the recipe
+  resolves the teacher's `deployment_shape` to the student deployment's shape
+  (`service.deployment_shape`) rather than leaving it unset. Heterogeneous
+  teachers on a different model than the student need an explicit shape —
+  pick one with `firectl deployment-shape list`. The recipe never creates
+  teacher deployments without a shape: shapeless deployments are the most
+  common cause of failed deployment creations, and the unshaped path may be
+  deprecated in the future.
 - Per-teacher metrics use `teacher_route/<slug>/scored` and
   `teacher_route/<slug>/inflight`; skewed datasets can leave some teacher
   deployments underused.
--- a/skills/fireworks-training/references/sdk-shapes.md
+++ b/skills/fireworks-training/references/sdk-shapes.md
@@ -37,11 +37,15 @@ trainer/deployment provisioning path.
 
 ## Deployment shape
 
-**Every dedicated deployment must use a deployment shape** — shapeless
-deployments skip validated configuration and fail far more often at creation
-(wrong GPU count, incompatible context length, unsupported quantization).
+**Do not create deployments without a shape.** A deployment is shapeless
+whenever the shape is omitted — whether or not accelerator fields are set —
+and a shapeless deployment skips validation (mistakes like a GPU count that
+can't fit the model surface only at creation, where they cause failures).
+Deployments created without a shape are the most common cause of failed
+deployment creations, and the unshaped path may be deprecated in the future.
 See the user-facing docs:
 [What is a deployment shape?](https://docs.fireworks.ai/faq-new/deployment-infrastructure/what-is-a-deployment-shape)
+If no shape fits a workload, contact Fireworks to find or add one.
 
 Do not set `cfg.deployment.deployment_shape` manually. The SDK resolves it from
 the requested deployment shape or the selected training profile, and recipes read
@@ -55,8 +59,10 @@ deployment_shape = service.deployment_shape
 
 That is a **versioned** path (`accounts/fw/deploymentShapes/ds-x/versions/abc123`).
 The `to_deployment_config` helper in `training/utils/config.py` **rejects
-deployments with no shape** and never forwards manual accelerator fields —
-the shape owns accelerator selection.
+deployments with no shape** (raising with guidance on how to resolve one).
+Overriding individual fields on top of a shape — e.g. `replica_count` — is
+supported; the shape owns accelerator selection, so manual accelerator fields
+are never forwarded.
 
 ## Reference-model shape (RL / DPO)
 
@@ -81,8 +87,8 @@ firectl deployment-shape list    # alias: firectl ds list
 
 When a customer needs a serving deployment shape for a model, list the
 model's verified shapes first (firectl is gaining richer per-model shape
-visibility) rather than guessing a GPU count — always create deployments
-against a listed shape.
+visibility, e.g. `firectl deployment-shape list`) rather than guessing a GPU
+count — always create deployments against a listed shape.
 
 Or programmatically via `FireworksClient` — see the SDK docs linked from the repo README.
 
--- a/skills/fireworks-training/references/sdk-shapes.md
+++ b/skills/fireworks-training/references/sdk-shapes.md
@@ -37,6 +37,12 @@ trainer/deployment provisioning path.
 
 ## Deployment shape
 
+**Every dedicated deployment must use a deployment shape** — shapeless
+deployments skip validated configuration and fail far more often at creation
+(wrong GPU count, incompatible context length, unsupported quantization).
+See the user-facing docs:
+[What is a deployment shape?](https://docs.fireworks.ai/faq-new/deployment-infrastructure/what-is-a-deployment-shape)
+
 Do not set `cfg.deployment.deployment_shape` manually. The SDK resolves it from
 the requested deployment shape or the selected training profile, and recipes read
 the resolved value from the service:
@@ -48,8 +54,9 @@ deployment_shape = service.deployment_shape
 ```
 
 That is a **versioned** path (`accounts/fw/deploymentShapes/ds-x/versions/abc123`).
-The `to_deployment_config` helper in `training/utils/config.py` auto-clears
-manual accelerator fields whenever a shape is present.
+The `to_deployment_config` helper in `training/utils/config.py` **rejects
+deployments with no shape** and never forwards manual accelerator fields —
+the shape owns accelerator selection.
 
 ## Reference-model shape (RL / DPO)
 
@@ -71,6 +78,11 @@ The CI pattern for the saves-GPUs variant is `ref_shape = "" if lora_rank > 0 el
 firectl training-shape list      # alias: firectl ts list
 firectl deployment-shape list    # alias: firectl ds list
 ```
+
+When a customer needs a serving deployment shape for a model, list the
+model's verified shapes first (firectl is gaining richer per-model shape
+visibility) rather than guessing a GPU count — always create deployments
+against a listed shape.
 
 Or programmatically via `FireworksClient` — see the SDK docs linked from the repo README.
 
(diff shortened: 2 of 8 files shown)


skills/fireworks-training/references/sdk-distillation.md
```diff
@@ -170,7 +170,11 @@ cfg.multi_teacher = MultiTeacherConfig(
   deployments.
 - `teacher_replica_count` controls replicas for auto-created frozen teacher
   deployments. `teacher_deployment_shape` sets the run-level default; individual
-  `TeacherConfig.deployment_shape` can override it.
+  `TeacherConfig.deployment_shape` can override it. When neither is set,
+  teachers default to the resolved student deployment shape — set one
```
greg-fw: Is this a statement of fact or a request? I don't think we have anything that makes them default, so make clear whether (a) we want the user to do something different or (b) we need to update our own code somehow to enforce this.

greg-fw: Statement of fact — (b), the code change. This same PR adds the enforcement in `training/recipes/distillation_loop.py`: `_resolve_teacher_runtime` reads `service.deployment_shape` and `_teacher_deployment_shape_for_spec` falls back to it when neither per-teacher nor run-level shape is set (tested in `test_teacher_deployment_shape_defaults_to_student_shape`). The doc text was ambiguous about that, so a18f137 rewrites it as enforced behavior with the resolution order and a pointer to the recipe code.

training/recipes/distillation_loop.py
```diff
@@ -314,11 +316,15 @@ def _validate_teacher_tokenizers(
 def _teacher_deployment_shape_for_spec(
     spec: TeacherConfig,
     cfg: Config,
+    *,
+    student_deployment_shape: str | None = None,
```
greg-fw: this feels like a very ugly addition to this existing API. What are our other options here? Should this be part of cfg? why/why not?

greg-fw: Also, why are we bothering with student deployment shape here?  Seems like an awkward thing, esp bc I'm not certain student and teacher deployments are likely to be the same type of model at all.

training/utils/config.py
```diff
@@ -311,8 +307,16 @@ def to_deployment_config(
         infra: InfraConfig,
     ) -> DeploymentConfig:
         """Produce an SDK-level DeploymentConfig from cookbook settings."""
-        skip_validation = False
```
greg-fw, line 314: in what cases would we possibly want to keep this?

Does this or the raise ValueError below possibly break existing use cases?

training/recipes/distillation_loop.py
greg-fw: Agreed it's ugly — a resolved-runtime value threaded through a private helper as a kwarg is the worst of both worlds: not config, but also not self-contained. Options I see:

1. **Keep the fallback, move it onto cfg** — e.g. resolve `student_deployment_shape` into `cfg.teacher_deployment_shape` once in `main()` before teacher specs are resolved, and keep the helper signature clean. Downside: `Config` is a user-facing dataclass; mutating it with a resolved runtime value blurs "user sets this" vs "recipe computes this."
2. **Drop the fallback, raise instead** (my lean, see my reply on your other comment): if neither per-teacher nor run-level shape is set, raise with guidance to pick a shape via `firectl deployment-shape list`. No new plumbing at all, consistent with `to_deployment_config`'s policy.
3. Keep as-is if you think the same-family default earns its complexity.

If you want the default kept, I'd do (1). If you agree with the other comment, (2) makes this moot.

greg-fw: Fair challenge. My reasoning was that the single-teacher path docstrings say `teacher_model` is 'on the same tokenizer as `base_model`' — i.e. the common case is same-family, where the student's shape is usually right. But you're correct that `multi_teacher` exists precisely for heterogeneous teachers, and there the student's shape is an actively wrong guess (would provision, e.g., an 8B student's shape for a 235B teacher — or worse, silently succeed with degraded serving).

Proposal: drop the student-shape fallback and **raise** when neither per-teacher `TeacherConfig.deployment_shape` nor run-level `teacher_deployment_shape` is set — same policy as `to_deployment_config`, message points at `firectl deployment-shape list` and contact-us-if-none-fits. This also makes your API-shape comment moot (no kwarg threading). Trade-off: unconfigured runs that previously silently created a shapeless deployment now fail fast at setup — which is the RFC-preferred outcome anyway, but it is a breaking change for anyone relying on the old silent path, so flagging it for your call. Happy to implement either way.

training/utils/config.py
greg-fw: **Keep cases for `to_deployment_config`:** I found none — zero non-test callers in the repo (recipes/examples/provision/case-studies/tito all use the SDK-managed service path via `service.py`), and no mention in the public README or llms.txt. It's a compat surface only.

**Breakage check on both changes:**
- `to_deployment_config` raise: cannot break any in-repo use case (no callers). Only risk is an external customer having written their own code against this helper — unknowable from here, but it was never documented publicly, so exposure should be minimal. It's also the documented-behavior escape: anyone who hits it gets a message telling them exactly how to fix it.
- Removing `deployment_accelerator_type`: same analysis — no in-repo readers; anyone constructing `DeployConfig(deployment_accelerator_type=...)` now gets a `TypeError`. Same exposure caveat.

If you'd rather not hard-remove the field yet, the softer option is a deprecation shim: keep the field, emit a `DeprecationWarning`, and ignore its value (shape-only). Given the RFC's trajectory (API-level blocking coming), I'd just delete — but say the word and I'll do the shim instead.

training/recipes/distillation_loop.py
greg-fw: Went with **option C** (b2ad2a7): dropped the `student_deployment_shape` kwarg entirely. When no explicit shape is configured, the recipe auto-selects the latest **validated** deployment shape for each teacher's own model via `_auto_select_teacher_deployment_shape` (one list call per unique teacher model, cached, logged at INFO) — mirroring trainer-side `auto_select_training_shape`. Raises with contact-us guidance only if a teacher model has zero validated shapes (a platform gap, not a user error). So the ugly kwarg is gone and nothing is blocked at setup.

greg-fw: You were right — student shape was a bad guess, especially for multi_teacher. Replaced in b2ad2a7 with per-teacher-model auto-selection: the recipe picks the latest validated deployment shape for **each teacher's own model**, not the student's. Blocked-user analysis (per your framing): experts unaffected (explicit shape wins); unconfigured users get a validated default for the right model instead of either a silent shapeless create or a wrong student shape; the only blocked case is a teacher model with no validated shapes at all, which fails fast with 'contact Fireworks to find or add one' — that's a platform gap, not something the user can self-heal by picking GPUs. Heterogeneous teachers now get the correct shape automatically rather than needing to know to configure one.

training/utils/config.py
greg-fw: Follow-up per the design decision: kept the hard delete (no deprecation shim) for `deployment_accelerator_type` and the raise in `to_deployment_config` — zero non-test callers verified, and the recipe path that actually runs (SDK-managed service via `service.py`) is unaffected. The distillation teacher path now auto-resolves validated shapes instead of raising (b2ad2a7), so the only remaining raise in this PR is on a helper with no callers plus the no-validated-shape-exists case, which is a contact-us failure by design.

skills/fireworks-training/references/sdk-distillation.md
```diff
@@ -169,8 +169,16 @@ cfg.multi_teacher = MultiTeacherConfig(
   region used by the SDK-managed student sampler and auto-created teacher
   deployments.
 - `teacher_replica_count` controls replicas for auto-created frozen teacher
-  deployments. `teacher_deployment_shape` sets the run-level default; individual
-  `TeacherConfig.deployment_shape` can override it.
+  deployments. Teacher shape resolution (enforced by the recipe code in
```
greg-fw: Sounds like you made this comment less readable and longer, even though AFAICT the first sentence above says the sam ething (more clearly) than the multiple sentences below.

The stuff from "The recipe never" seems useful, though. "shapeless / unshaped" -> pick one. Say "shapeless deployments" and then "shapeless path". Remove the contact fireworks stuff.

skills/fireworks-training/references/sdk-shapes.md
```diff
@@ -37,6 +37,16 @@ trainer/deployment provisioning path.
 
 ## Deployment shape
 
+**Do not create deployments without a shape.** A deployment is shapeless
```
greg-fw: "A deployment is shapeless whenever...where they cause failures)" remove all this, it's extra cruft.

Remove the "user-facing docs" shit. just make the "shape" a link to that.

Make the contact an actual email address, and remove the "add one part"

skills/fireworks-training/references/sdk-shapes.md
```diff
@@ -48,8 +58,11 @@ deployment_shape = service.deployment_shape
 ```
 
 That is a **versioned** path (`accounts/fw/deploymentShapes/ds-x/versions/abc123`).
-The `to_deployment_config` helper in `training/utils/config.py` auto-clears
-manual accelerator fields whenever a shape is present.
+The `to_deployment_config` helper in `training/utils/config.py` **rejects
+deployments with no shape** (raising with guidance on how to resolve one).
```
greg-fw, line 54: make "shape" a link to docs

skills/fireworks-training/references/sdk-shapes.md
```diff
@@ -72,6 +85,11 @@ firectl training-shape list      # alias: firectl ts list
 firectl deployment-shape list    # alias: firectl ds list
 ```
 
+When a customer needs a serving deployment shape for a model, list the
```
greg-fw: wtf is this?? "a customer"?? is this a note to the user or to ourselves? Looks like crap.

skills/fireworks-training/references/sdk-distillation.md
greg-fw: Fixed in e381432 — restored the original concise first sentence (`teacher_deployment_shape` run-level default, per-teacher override), then one sentence for the new auto-select behavior and one for the never-shapeless warning. One term throughout: 'shapeless deployments' / 'shapeless path'. Contact-Fireworks line removed.

skills/fireworks-training/references/sdk-shapes.md
greg-fw: Fixed in e381432 — the whole 'A deployment is shapeless whenever...' block is gone; the section is now two sentences plus the existing how-to. 'shape' is now a link to the docs page on first mention (no separate user-facing-docs line), and the contact is an email address with the 'add one' part removed.

skills/fireworks-training/references/sdk-shapes.md
greg-fw: Done — 'shape' is now linked to the docs page there too.

skills/fireworks-training/references/sdk-shapes.md
greg-fw: Yeah, that was a note-to-self masquerading as docs. Deleted; replaced with a single user-facing line: 'Always create deployments against a listed shape — never guess a GPU count.'

training/recipes/distillation_loop.py
```diff
@@ -315,13 +321,50 @@ def _teacher_deployment_shape_for_spec(
...
+    """Auto-select the latest validated deployment shape for a teacher model.
+
+    Deployments created without a shape are the most common cause of failed
+    deployment creations, so teacher deployments are never created shapeless.
+    Mirrors the trainer-side ``auto_select_training_shape`` pattern: pick from
+    what the control plane has already validated for this model.
+    """
+    from urllib.parse import urlencode
+
+    parent = f"{teacher_model.rsplit('/models/', 1)[0]}/deploymentShapes/-"
```
greg-fw: why are we re-writing auto_select_training_shape instead of just re-using it?

greg-fw: Fair — done in 12acdc3. `_auto_select_teacher_deployment_shape` now calls the existing `auto_select_training_shape` (exported from `training.utils`) for the teacher's model and takes `resolve_training_profile(...).deployment_shape` from the result — the same public two-step every other caller uses. The hand-rolled `deploy_mgr._get` listing and the `DeploymentManager` construction are gone; `main()` passes a `TrainerJobManager` instead. Also kills the private-SDK-method coupling I'd flagged earlier. New failure mode (profile resolves but has no pinned deployment shape) raises with guidance — that's a broken shape link, consistent with the verifier's treatment of the same case.

skills/fireworks-training/references/sdk-shapes.md
```diff
@@ -72,6 +77,8 @@ firectl training-shape list      # alias: firectl ts list
 firectl deployment-shape list    # alias: firectl ds list
 ```
 
+Always create deployments against a listed [shape](https://docs.fireworks.ai/faq-new/deployment-infrastructure/what-is-a-deployment-shape) — never guess a GPU count.
```
greg-fw: remove the "never guess" part.

skills/fireworks-training/references/sdk-shapes.md
```diff
@@ -72,6 +77,8 @@ firectl training-shape list      # alias: firectl ts list
 firectl deployment-shape list    # alias: firectl ds list
```
greg-fw: This command AFAICT does not exist.

skills/fireworks-training/references/sdk-shapes.md
greg-fw: Removed in 8a2a21c — that line is now 'Always create deployments against a shape returned by `match`.'

skills/fireworks-training/references/sdk-shapes.md
greg-fw: You were right, and it was worse than a wrong alias: `deployment-shape list` only exists inside `firectl-admin` (registered behind the IsAdmin gate in firectl.go), so public firectl rejects it — which matches the 'unknown command' you got locally. Fixed in 8a2a21c: the section now leads with `firectl deployment-shape-version match --model <MODEL>` (from fw-ai/fireworks#48393, with `--enable-addons` from #49060) — the server-side RPC that returns only shapes the account can actually deploy — and keeps `list --base-model` only as a raw-inventory note. Also swapped the same guidance in training/README.md and the to_deployment_config error message, and removed the last shapeless fallbacks in deploy-and-troubleshoot.md.

